PR #2, opened 2022-11-09
Fixes for `no-std` build

mkeeter:
- Fix Serde dependency to build with `no_std`
- Remove `ids_from_logical_ports`, since Ben was already planning to delete it
- Change the one remaining function that returned a `Vec` -> returning an iterator instead
- Drop `heapless` dependency, since we don't need it anymore
- Update tests
- Make `port_to_ids` public, since otherwise no one is using it (?)

### DIFF
--- a/messages/Cargo.toml
+++ b/messages/Cargo.toml
@@ -15,11 +15,6 @@ version = "1"
 features = [ "derive" ]
 default-features = false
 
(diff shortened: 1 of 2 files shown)


bnaecker (approved):
👍 Thanks!

PR #5, opened 2022-11-09
Adds mechanisms to specify which management interface a transceiver uses

bnaecker:
- Adds support for declaring the management interface specification a transceiver conforms to. Currently supports SFF-8636 and CMIS, all others are unknown and unsupported.
- Adds better mechanisms for addressing the upper page of a transceiver memory map. It now supports an addressing scheme per-specification, which allows the older single-level paged memory of SFF-8636, and the newer page-and-bank-select mechanisms in CMIS. All values are checked against the spec, for both pages and banks, though that doesn't validate against the pages supported by the modules themselves.
- More tests for above.

### DIFF
--- a/messages/src/lib.rs
+++ b/messages/src/lib.rs
@@ -59,9 +59,11 @@ pub enum Error {
     /// A request would result in a response that is too large to fit in a
     /// single UDP message.
     RequestTooLarge,
-    /// Someone sent an unexpected message (e.g. the host sending an SpRequest)
+
+    /// Someone sent an unexpected message (e.g. the host sending an SpRequest).
     ProtocolError,
-    /// The version in the header is unexpected
+
+    /// The version in the header is unexpected.
     VersionMismatch,
 }
 
--- a/messages/src/lib.rs
+++ b/messages/src/lib.rs
@@ -65,8 +65,16 @@ pub enum Error {
     VersionMismatch,
 }
 
-type MaskType = u32;
+// We're currently only expecting to be able to address up to 16 transceivers
+// per Sidecar FPGA. The `PortMask` below exposes this publicly, so we may wish
+// to hide it if / when we want to support other front I/O board designs with
+// different arrangements of FPGAs and / or transceivers.
+type MaskType = u16;
 
+/// A bitmask used to identify the set of transceiver ports to which a message
+/// applies.
+///
+/// Note that this bitmask is always per-FPGA.
 #[derive(Clone, Copy, Debug, Default, Deserialize, PartialEq, Serialize, SerializedSize)]
 pub struct PortMask(pub MaskType);
 
@@ -184,7 +192,7 @@ mod tests {
 
     #[test]
     fn test_port_mask_test_set_clear() {
-        let mut mask = PortMask(0b101u32);
+        let mut mask = PortMask(0b101);
         assert!(mask.is_set(0).unwrap());
         assert!(!mask.is_set(1).unwrap());
         assert!(mask.is_set(2).unwrap());
--- a/messages/src/lib.rs
+++ b/messages/src/lib.rs
@@ -150,24 +150,6 @@ pub struct ModuleId {
     pub ports: PortMask,
 }
 
-// Notes:
-//
-// Address 1 FPGA at a time in each message, a const-sized bitmask (u32? u64?)
-// for the ports on it.
-//
-// Dendrite knows how to go from logical port to (fpga_id, bit), based on the
-// front IO board identity.
-//
-// Need message to tell SP the kind of device in any given port. Really it's
-// which spec it conforms to, so that it can go read the temperature correctly
-// from the memory map and plug that into its thermal loop.
-//
-// Be aware that a read for more than one device returns a uint8_t ** -- I.e.,
-// we get an array of buffers, one for the data read in for each device.
-//
-// Same thing for a write, unless we want to only support writing the same data
-// to every port.
-
 #[cfg(test)]
 mod tests {
     use super::Error;
--- a/messages/src/lib.rs
+++ b/messages/src/lib.rs
@@ -2,11 +2,15 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
+// Copyright 2022 Oxide Computer Company
+
 #![cfg_attr(all(not(test), not(feature = "std")), no_std)]
 
-//! Messaging formats for managing the Sidecar QSFP ports over the network.
+//! Messaging formats for managing the Sidecar transceiver ports over the
+//! network.
 
 pub mod message;
+pub mod mgmt;
 
 use hubpack::SerializedSize;
 use serde::Deserialize;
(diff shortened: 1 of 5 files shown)


messages/src/mgmt/mod.rs
```diff
@@ -0,0 +1,222 @@
...
+impl MemoryRegion {
+    /// Construct a new memory region.
+    ///
+    /// Note that you must specify an `upper_page`, even if you are only
+    /// accessing the fixed, lower page of memory. In that case, you can use
+    /// the default for the specification of the accessed module, which uses
+    /// page and / or bank zero.
+    pub fn new(upper_page: UpperPage, offset: u8, len: u8) -> Result<Self, Error> {
+        // The last accessed byte must be within the 256-byte memory map.
+        if offset.checked_add(len).is_none() {
```
mkeeter, line 154: There may be an off-by-one issue here:

If `offset = 250` and `len = 6`, we _should_ be able to read bytes 250-255 (inclusive), but `checked_add` will give `None` (and hence this will return an error).

bnaecker: Ah, nice catch, thanks. I was trying to avoid a bunch of conversions or casts, but I think that might actually be the clearest mechanism. I'll sort it, and add a regression. Thank you.

messages/src/mgmt/mod.rs
```diff
@@ -0,0 +1,222 @@
...
+}
+
+/// A description of a region of a transceiver's memory map.
+///
+/// See [`UpperPage`] for a detailed description of how a module's upper page is
+/// mapped in for an operation. This type also performs validation against the
+/// offset / length, to ensure they're a valid access for the 256-byte memory
+/// map of all transceiver modules we support.
+#[derive(Clone, Copy, Debug, Deserialize, PartialEq, Serialize, SerializedSize)]
+pub struct MemoryRegion {
```
mkeeter, line 139: General question: if the SP receives a `Read(MemoryRegion)` command over the network, is _it_ responsible for handling the memory region (e.g. configuring pages and banks)?

That's what's implied by our API, but I'm not sure whether the BF SDE expects that to be happening, or whether the SDE wants to manage pages itself (i.e. through manual reads and writes to `0x7E` and `0x7F`).

messages/src/mgmt/mod.rs
bnaecker: Resolved in https://github.com/oxidecomputer/transceiver-control/pull/5/commits/9aa5afd9d3e88eeab9164143d05a957d56588650

messages/src/message.rs
```diff
@@ -215,33 +169,16 @@ bitflags::bitflags! {
...
+    /// Entirely power off a transceiver module.
+    Off,
+    /// Request the module transition into low-power mode.
+    Low,
+    /// Request the module transition into high-power mode.
+    ///
+    /// Note that the actual power mode is defined in the transceiver's memory
+    /// map, in a location that depends on the exact management specification it
+    /// conforms to.
+    High,
```
mkeeter, line 183: This API seems to leave a lot up to the SP.

For example, with an SFF-8636 device, we may need to set extra bits depending on class (class 5-7 need bit 2 in B93 set; class 8 need bit 3 in B93).

I'm confused, because it seems like we don't want the SP to have this much responsibility for parsing transceiver types.

Maybe the API is only intended to control the `LPMode` pin?  If so, we should give it a name that reflects the intention of controlling just the GPIO, instead of `PowerMode` more generally.

messages/src/mgmt/mod.rs
bnaecker: That's correct. I should make that clearer in the message definitions.

It's really just about atomicity from the point of view of the host. If we make those separate requests, then theoretically we could interleave two separate requests that set the page, and neither would be accessing the correct data.

The question about SDE integration is a good one. It's not clear exactly how they use it, but they do expect to be able to write some register, potentially the page-select, and then read something else from that selected page. My plan to deal with that is to cache the requested page that the SDE thinks it's setting, and then always send it along in the request to the SP to ensure it's atomic.

Hopefully that clears things up, but I could very well be missing other considerations.

messages/src/lib.rs
```diff
@@ -108,106 +129,61 @@ impl PortMask {
...
-    const STRIDE: u8 = 8;
-    const PORTS_PER_FPGA: u8 = 16;
-    const NUM_FPGAS: u8 = 2;
-    let fpga_id = (port / STRIDE) % NUM_FPGAS;
-    let port_index = ((port / PORTS_PER_FPGA) * STRIDE) + (port % STRIDE);
-    Fpga::try_from(fpga_id).map(|fpga| (fpga, port_index))
-}
+// Notes:
+//
+// Address 1 FPGA at a time in each message, a const-sized bitmask (u32? u64?)
```
mkeeter, line 155: `u16`?

bnaecker: Lol, this should just go away those were my own notes.

messages/src/message.rs
bnaecker: > This API seems to leave a lot up to the SP.

That's fair, and definitely something we're trying to avoid! Yes, we don't want the SP mucking much with the memory map to determine the transceiver type.

I was expecting this to be effectively control of the LPMode pin, plus whatever the FPGA needs to do to cut the power entirely. If we wanted to set the actual power class, we'd do that in a preceding request. That may run afoul of the atomicity we're trying to achieve though. It might be that we need a compound message or something, which instructs the SP _how_ to set the power class, and also enable high-power mode? I.e., which bytes to write, and please de-assert LPMode for me.

Does that sound better? I think the message itself could just be a `MemoryRegion` inside that variant, but I may be missing something.

messages/src/lib.rs
bnaecker: Removed in `c8233ff`

messages/src/mgmt/mod.rs
mkeeter: This kind of stateful system is always tricky, but I think the plan sounds reasonable.  On the SP side, I'll look at caching configured page and banks, so that we don't do a bunch of extra transactions.

bnaecker: I think by "extra transactions", do you mean to avoid an I2C write to set the page-select, if we're not changing the page from the previous read? I think that should be OK, but maybe unnecessary. I have basically zero idea what the frequency of these messages will be, so an argument could be made to defer that. Whatever you think is best!

Would more documentation be helpful, either on these variants or a crate comment?

messages/src/message.rs
mkeeter: My vague intuition would be to split "go to high-power mode" into two separate messages:
- Configure the relevant registers
- Deassert LPMode

(since the registers have no effect until the pin is low)

I guess I'm not sure what we're trying to protect against with the atomic request strategy.  As long as the host doesn't allow anyone else to inject messages midway through the "go to high power" process, I don't know who else would be modifying transceiver configuration.

bnaecker: I've been looking at the specs in more detail, and you may be right that sending two messages is fine. I was under the impression that there may be a complex sequence of power modes that modules go through during initialization or something, and that the SDE would need to be sure that those have taken effect before deasserting `LPMode`. On closer reading, I see it's really about limiting the maximum power of a free-side module, in case some other, older component in the system would be damaged by the later power class 8, or the system isn't capable of supplying that power at all.

That is, we're generally just going to set byte 93 bits 2/3 for SFF-8636, and then deassert `LPMode` when we're ready. There's no real reason to expect that the SDE would separately twiddle those bits in a different way from the control plane. For CMIS modules, it doesn't look like there is a concept of limiting the power, just that there _is_ a high-power mode. (I could be missing something.)

You brought up earlier whether this API is intended to control just the `LPMode` pin. I still think we want to be able to completely power down a module, but I'm open to how we actually express that in the API. It seems like this would be sufficient, combined with a _separate_ write to 93[2-3] where needed (SFF-8636 modules). Does that sound reasonable? Or would you prefer separate messages for controlling the `LPMode` pin and completely powering down a module?

As a related note, I think we'll need to update the status bitflags I've already got. There's no way to express a module being present, but completely powered down, I think. Does that sound right?

bnaecker: Ah, no my point about the status flags is not correct. The `Status::ENABLED` bit is outside the spec, and tells us whether there is power at all delivered to the module, independent of whether low- or high-power mode is enabled. Nevermind!

mkeeter: Thanks for the details!  After thinking a little more, I think this `enum` would be fine with some additional commenting, basically saying

> - In `Off` mode, we cut power to the transceiver using the hot-swap / e-fuse
> - In `Low` mode, we enable power to the transceiver but keep `LPMode` high.  Please do not set `Power_override`, as this defeats the purpose of controlling this in hardware!
> - In `High` mode, we enable power to the transceiver and pull `LPMode` low; further configuration of registers may be required to actually enter high-power mode (depending on device class)

bnaecker: Thanks, that sounds good! I believe I've addressed this in https://github.com/oxidecomputer/transceiver-control/pull/5/commits/4575e0e1220151a1208842c81bba667cf6dd64fe, but let me know if you have anything outstanding.

rmustacc: FWIW, when I originally wrote this, my assumption was that the host would configure any of the actual registers itself before asking the SP to the release the low power pin. So my expectation was that the onus was not on the SP.

mkeeter (approved):

messages/src/message.rs
bnaecker: Yep, understood! I think @mkeeter and I agree on that point as well. The bulk of this discussion was about whether writing the relevant registers and releasing into high-power mode needed to be a single message from host to SP. I think we arrived at "no", at least for the time being.

Part of the reason is that the host library I'm writing will include a function to enable high-power mode. That will use two exchanges under the hood, to set the correct registers if appropriate, and then deassert `LPMode`, but we _think_ that should be OK at this point. Let me know if you think that's _not_ the case @rmustacc. In the meantime, I'll make that assumption clear in the documentation.

rmustacc: I believe that should be fine. What I was trying to avoid was what seemed like at one point an additional command that was specific to this process of setting the i2c registers.

PR #11, opened 2022-11-15
Makes reads/writes of transceiver modules size-limited

bnaecker:
- Splits the `MemoryRegion` type into `Memory{Read,Write}`. These have different size limits.
- Adds the `MemoryPage` trait, implemented for the pages we currently support (SFF-8636 and CMIS). This describes the start / end of the page, as well as the maximum size of a read or write. These are used to construct the `Memory{Read,Write}` types for a specific page, where the offset / length of the access are validated.
- Beefs up tests.

### DIFF
--- a/messages/src/mgmt/mod.rs
+++ b/messages/src/mgmt/mod.rs
@@ -26,8 +26,7 @@ pub enum ManagementInterface {
     Unknown(u8),
 }
 
-/// A description of how to page in the upper 128-bytes of a transceiver's
-/// memory map.
+/// A description of a single 128-byte page of a transceiver's memory map.
 ///
 /// There are several specifications that define the memory map for a free-side
 /// transceiver monitor. (See [`ManagementInterface`] for details.) Though there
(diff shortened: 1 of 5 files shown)


messages/src/mgmt/mod.rs
```diff
@@ -87,145 +86,317 @@ pub enum ManagementInterface {
...
+// Helper to check that the limits provided define a valid read.
+fn verify_read<P>(page: &P, offset: u8, len: u8) -> Result<(), Error>
+where
+    P: MemoryPage,
+{
+    if offset >= page.min_offset()
+        && offset <= page.max_offset()
+        && (offset as u16 + len as u16) <= page.upper_limit()
+        && len >= P::MIN_READ_SIZE
+        && len <= P::MAX_READ_SIZE
```
mkeeter, line 254: I stared at this a lot, and can't find any off-by-one errors!

bnaecker: There's a first time for everything :)

mkeeter (approved):
Looks great!

PR #13, opened 2022-11-16
Adds basic host-side controller.

bnaecker:
- Adds a `Controller`, the main type which handles the network protocol between the host and SP. This spawns a tokio task for managing the actual UDP socket and for calling the user-supplied request-handler function. These communicate with the main controller via channels. The public API is basically a wrapper converting the data for each function into the right RPC message, and then passing that to the network task for sending and receiving the response.
- Adds a small CLI for running the controller manually. It currently does very little.
- Adds a dumb mock SP, which just responds to memory reads with zeros of the right size.

### DIFF
--- a/controller/src/bin/mock-sp.rs
+++ b/controller/src/bin/mock-sp.rs
@@ -9,15 +9,15 @@ use transceiver_messages::message::MessageBody;
 use transceiver_messages::message::SpResponse;
 use transceiver_messages::MAX_PAYLOAD_SIZE;
 use transceiver_messages::PORT;
-
-const PEER: Ipv6Addr = Ipv6Addr::new(0xff02, 0, 0, 0, 0, 0, 0x1de, 2);
+use transceiver_messages::ADDR;
 
 #[tokio::main]
 async fn main() {
+    let peer = Ipv6Addr::from(ADDR);
     let sock = UdpSocket::bind(SocketAddrV6::new(Ipv6Addr::UNSPECIFIED, PORT, 0, 0))
         .await
         .unwrap();
-    sock.join_multicast_v6(&PEER, 0).unwrap();
+    sock.join_multicast_v6(&peer, 0).unwrap();
 
     let mut buf = [0; MAX_PAYLOAD_SIZE * 2];
     loop {
--- a/controller/src/bin/mock-sp.rs
+++ b/controller/src/bin/mock-sp.rs
@@ -1,0 +1,55 @@
+//! Mock SP server that dummies up basic QSFP memory map.
+
+use std::net::Ipv6Addr;
+use std::net::SocketAddrV6;
+use tokio::net::UdpSocket;
+use transceiver_messages::message::HostRequest;
+use transceiver_messages::message::Message;
+use transceiver_messages::message::MessageBody;
+use transceiver_messages::message::SpResponse;
+use transceiver_messages::MAX_PAYLOAD_SIZE;
+use transceiver_messages::PORT;
+
+const PEER: Ipv6Addr = Ipv6Addr::new(0xff02, 0, 0, 0, 0, 0, 0x1de, 2);
+
+#[tokio::main]
+async fn main() {
+    let sock = UdpSocket::bind(SocketAddrV6::new(Ipv6Addr::UNSPECIFIED, PORT, 0, 0))
+        .await
+        .unwrap();
+    sock.join_multicast_v6(&PEER, 0).unwrap();
+
+    let mut buf = [0; MAX_PAYLOAD_SIZE * 2];
+    loop {
+        match sock.recv_from(&mut buf).await {
+            Err(e) => println!("{e:?}"),
+            Ok((n_bytes, peer)) => {
+                let (message, _) = hubpack::deserialize::<Message>(&buf[..n_bytes]).unwrap();
+                println!("=> {message:?}");
+                match message.body {
+                    MessageBody::HostRequest(HostRequest::Read(read)) => {
+                        let response = Message {
+                            header: message.header,
+                            modules: message.modules,
+                            body: MessageBody::SpResponse(SpResponse::Read(read)),
+                        };
+                        // Flat array of data.
+                        let data = vec![
+                            0;
+                            usize::from(read.len())
+                                * message.modules.ports.0.count_ones() as usize
+                        ];
+                        println!("data size: {}", data.len());
+                        println!("<= {response:?}");
+                        let n_bytes = hubpack::serialize(&mut buf, &response).unwrap();
+                        buf[n_bytes..n_bytes + data.len()].copy_from_slice(&data);
+                        sock.send_to(&buf[..n_bytes + data.len()], &peer)
+                            .await
+                            .unwrap();
+                    }
+                    _ => {}
+                }
+            }
+        }
+    }
+}
--- a/controller/src/bin/mock-sp.rs
+++ b/controller/src/bin/mock-sp.rs
@@ -7,9 +7,9 @@ use transceiver_messages::message::HostRequest;
 use transceiver_messages::message::Message;
 use transceiver_messages::message::MessageBody;
 use transceiver_messages::message::SpResponse;
+use transceiver_messages::ADDR;
 use transceiver_messages::MAX_PAYLOAD_SIZE;
 use transceiver_messages::PORT;
-use transceiver_messages::ADDR;
 
 #[tokio::main]
 async fn main() {
--- a/controller/src/lib.rs
+++ b/controller/src/lib.rs
@@ -123,7 +123,8 @@ pub struct Controller {
     // The task handling actual network IO with the peer.
     io_task: JoinHandle<()>,
 
-    // The task handling actual network IO with the peer.
+    // The task receiving requests from the peer and calling the user-supplied
+    // request handler.
     request_task: JoinHandle<()>,
 }
 
--- a/controller/src/lib.rs
+++ b/controller/src/lib.rs
@@ -2,10 +2,605 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
+// Copyright 2022 Oxide Computer Company
+
 //! A host-side control interface to the SP for managing Sidecar transceivers.
 
+use hubpack::SerializedSize;
+use slog::debug;
+use slog::error;
+use slog::info;
+use slog::warn;
+use slog::Logger;
+use std::future::Future;
+use std::net::Ipv6Addr;
+use std::net::SocketAddr;
+use std::net::SocketAddrV6;
+use std::sync::atomic::AtomicU64;
+use std::sync::atomic::Ordering;
+use std::time::Duration;
 use tokio::net::UdpSocket;
+use tokio::sync::mpsc;
+use tokio::sync::oneshot;
+use tokio::task::JoinHandle;
+use tokio::time::sleep_until;
+use tokio::time::Instant;
+use transceiver_messages::message;
+use transceiver_messages::message::Header;
+use transceiver_messages::message::HostRequest;
+use transceiver_messages::message::HostResponse;
+use transceiver_messages::message::Message;
+use transceiver_messages::message::MessageBody;
+use transceiver_messages::message::Status;
+use transceiver_messages::mgmt::MemoryRead;
+use transceiver_messages::Error as MessageError;
+use transceiver_messages::ModuleId;
+use transceiver_messages::ADDR;
+use transceiver_messages::MAX_PAYLOAD_SIZE;
+use transceiver_messages::PORT;
+
+/// An error related to managing the transceivers.
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    #[error("Error in transceiver control protocol: {0:?}")]
+    Protocol(#[from] transceiver_messages::Error),
+
+    #[error("Network or I/O error: {0}")]
+    Io(#[from] std::io::Error),
+
+    #[error("Message type requires data, but none provided")]
+    MessageRequiresData,
+
+    #[error("A serialization error occurred: {0}")]
+    SerDes(hubpack::Error),
+}
+
+// A request sent from host to SP, possibly with trailing data.
+#[derive(Clone, Debug)]
+struct HostRpcRequest {
+    pub message: Message,
+    pub data: Option<Vec<u8>>,
+}
+
+// A response sent from SP to host, possibly with trailing data.
+#[derive(Clone, Debug)]
+#[allow(dead_code)]
+struct SpRpcResponse {
+    pub message: Message,
+    pub data: Option<Vec<u8>>,
+}
+
+/// A request sent from SP to host, possibly with trailing data.
+#[derive(Clone, Debug)]
+pub struct SpRpcRequest {
+    pub message: Message,
+    pub data: Option<Vec<u8>>,
+}
+
+/// A response sent from host to SP, possibly with trailing data.
+#[derive(Clone, Debug)]
+pub struct HostRpcResponse {
+    pub message: Message,
+    pub data: Option<Vec<u8>>,
+}
+
+// A request from host to SP that has not yet been completed.
+#[derive(Debug)]
+struct OutstandingHostRequest {
+    // The actual request object we're sending. It's stored so that we can
+    // resend it if needed.
+    request: HostRpcRequest,
+    // The channel on which the eventual reply will be sent.
+    response_tx: oneshot::Sender<Result<SpRpcResponse, Error>>,
+    // The time we last sent the request, used during retries.
+    last_sent: Instant,
+}
+
+// We limit ourselves to a single outstanding request in either direction at
+// this point.
+const NUM_OUTSTANDING_REQUESTS: usize = 1;
 
 /// A type for controlling transceiver modules on a Sidecar.
 #[derive(Debug)]
-pub struct Controller;
+pub struct Controller {
+    _log: Logger,
+    message_id: AtomicU64,
+
+    // Channel onto which requests from the host to SP are sent.
+    //
+    // `io_task` owns the receiving end of this, and actually sends out the
+    // messages to the SP.
+    outgoing_request_tx: mpsc::Sender<OutstandingHostRequest>,
+
+    // The task handling actual network IO with the peer.
+    io_task: JoinHandle<()>,
+
+    // The task handling actual network IO with the peer.
+    request_task: JoinHandle<()>,
+}
+
+impl Drop for Controller {
+    fn drop(&mut self) {
+        self.io_task.abort();
+        self.request_task.abort();
+    }
+}
+
+impl Controller {
+    /// Create a new transceiver controller.
+    ///
+    /// `request_handler` is a function that yields responses to SP requests. As
+    /// requests over the network are received, they'll be passed into the
+    /// handler, and the yielded response forwarded back to the SP.
+    pub async fn new<H, F>(log: Logger, request_handler: H) -> Result<Self, Error>
+    where
+        H: Fn(SpRpcRequest) -> F + Send + Sync + 'static,
+        F: Future<Output = Result<HostRpcResponse, Error>> + Send,
+    {
+        // TODO-correctness We probably want to accept a specific address as
+        // part of the construction of this object.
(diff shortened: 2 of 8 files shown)


bnaecker:
This is a MVP for the the host-side of the transceiver controller. It implements the core of the network communication, an has just enough of a public API to get a feel for how it works. There's also a mock SP which only responds to a request to read transceiver maps, with all zeros but of the correct size. There's also a small CLI `xcvradm`, which may or may not merge with Matt's draft of the demo CLI. The point of this one was just to exercise the `Controller` type.

This is still a bit of a rough draft (there are no tests, for example). I would love some feedback before building the rest of the public controller API though.

controller/src/lib.rs
```diff
@@ -2,10 +2,605 @@
...
+    // Channel onto which requests from the host to SP are sent.
+    //
+    // `io_task` owns the receiving end of this, and actually sends out the
+    // messages to the SP.
+    outgoing_request_tx: mpsc::Sender<OutstandingHostRequest>,
+
+    // The task handling actual network IO with the peer.
+    io_task: JoinHandle<()>,
+
+    // The task handling actual network IO with the peer.
```
mkeeter, line 118: Same docstring comment as above?

controller/src/lib.rs
```diff
@@ -2,10 +2,605 @@
...
+            io_task,
+            request_task,
+        })
+    }
+
+    // Return a header using the next available message ID.
+    fn next_header(&self) -> Header {
+        Header {
+            version: message::version::V1,
+            message_id: self.message_id.fetch_add(1, Ordering::SeqCst),
```
mkeeter, line 209: Are we using an atomic here for interior mutability, or do we also need `Sync`?

If we only need interior mutability, a `Cell<u64>` might better reflect intent, e.g.
```rust
        let message_id = self.message_id.take();
        self.message_id.set(message_id + 1);
        Header {
            version: message::version::V1,
            message_id,
        }
```

(alas, [`Cell::update`](https://doc.rust-lang.org/std/cell/struct.Cell.html#method.update) remains unstable)

controller/src/lib.rs
```diff
@@ -2,10 +2,605 @@
...
+// These are serialized and sent over the contained UDP socket to the multicast
+// address defined in `transceiver_messages::ADDR`.
+//
+// This task also listens for incoming messages on the UDP socket from the SP.
+// These are deserialized and sanity checked. (Obvious failures result in an
+// error being sent back immediately.) Assuming they seem reasonable, then they
+// are dispatched as follows:
+//
+// - Requests from the SP are sent to the request handler, via
+// `incoming_request_tx.
```
mkeeter, line 299:
```suggestion
// `incoming_request_tx`.
```

controller/src/lib.rs
bnaecker: My hunch is that we need `Sync`, but I'm not positive about that. The main reason is that I think this type is going to be shared between the `dpd` API server (really part of a `Switch` object it maintains), and also used by the BF SDE. The `Switch` type uses a `Mutex`, so accesses via `dpd` would be synchronized. But the BF SDE just makes FFI calls from C, so I'd expect that we to synchronize those.

controller/src/lib.rs
```diff
@@ -2,10 +2,605 @@
...
+    peer: &SocketAddrV6,
+    request: &mut OutstandingHostRequest,
+    buf: &mut [u8],
+) -> Result<usize, Error> {
+    let mut msg_size = match hubpack::serialize(buf, &request.request.message) {
+        Ok(n) => n,
+        Err(e) => return Err(Error::SerDes(e)),
+    };
+    if let Some(data) = &request.request.data {
+        buf[msg_size..data.len()].copy_from_slice(data);
```
bnaecker, line 580: There's a bug here. The end should be `msg_size + data.len()`.

controller/src/lib.rs
```diff
@@ -2,10 +2,605 @@
...
+                        );
+                        continue;
+                    }
+                    Ok((msg, remainder)) => (msg, remainder),
+                };
+                debug!(log, "message from peer"; "peer" => peer, "message" => ?message);
+
+                // Sanity check the protocol version.
+                if message.header.version != message::version::V1 {
+                    if let Err(e) = send_version_mismatch(&log, &socket, &peer, &message.header, &message.modules).await {
```
mkeeter, line 422: I think we only want to `send_version_mismatch` if we received a mismatched `HostRequest`; if we received a mismatched `HostResponse`, then something _else_ is wrong, but replying to the SP won't help.

controller/src/lib.rs
```diff
@@ -2,10 +2,605 @@
...
+                let (message, remainder): (Message, _) = match hubpack::deserialize(&recv_buf) {
+                    Err(e) => {
+                        error!(
+                            log,
+                            "failed to deserialize message";
+                            "reason" => ?e,
+                            "peer" => peer,
+                            "n_bytes" => n_bytes,
+                        );
+                        continue;
```
mkeeter, line 414: Per [this discussion](https://github.com/oxidecomputer/hubris/pull/931#discussion_r1022991928), we should try deserializing just the `Header` if deserializing the whole message fails.

In the case of the SP sending a `HostRequest` with a mismatched version, deserializing the header will let us use the correct `message_id` in our `Error` reply.

controller/src/lib.rs
```diff
@@ -2,10 +2,605 @@
...
+                let data = if expected_len > 0  {
+                    if remainder.len() < expected_len {
+                        warn!(
+                            log,
+                            "message did not contain expected data";
+                            "expected_len" => expected_len,
+                            "actual_len" => remainder.len(),
+                            "peer" => peer,
+                        );
+                        send_protocol_error(&log, &socket, &peer, &message.header, &message.modules).await;
```
mkeeter, line 454: Same question here: if the SP is _replying_ to us, do we need to inform it of this error?

controller/src/lib.rs
```diff
@@ -2,10 +2,605 @@
...
+    log: &Logger,
+    socket: &UdpSocket,
+    peer: &SocketAddr,
+    header: &Header,
+    modules: &ModuleId,
+) {
+    error!(log, "Protocol error");
+    let err = MessageError::ProtocolError;
+    let message = Message {
+        header: *header,
```
mkeeter, line 520: Nit: if we're going to deference them immediately, might as well pass `header` and `modules` in by value.

controller/src/lib.rs
```diff
@@ -2,10 +2,605 @@
...
+        actual: header.version,
+    };
+    let message = Message {
+        header: *header,
+        modules: *modules,
+        body: MessageBody::HostResponse(HostResponse::Error(err)),
+    };
+    let mut buf = [0u8; Message::MAX_SIZE];
+    hubpack::serialize(&mut buf, &message).unwrap();
+    let n_bytes = socket.send_to(&buf, peer).await?;
```
mkeeter, line 563: same here:
```suggestion
    let serialized_len = hubpack::serialize(&mut buf, &message).unwrap();
    let n_bytes = socket.send_to(&buf[..serialized_len], peer).await?;
```

controller/src/lib.rs
```diff
@@ -2,10 +2,605 @@
...
+    assert_eq!(n_bytes, Message::MAX_SIZE);
```
mkeeter, line 564: Once making the edits above, this would become
```rust
assert_eq!(n_bytes, serialized_len);
```

controller/src/lib.rs
```diff
@@ -2,10 +2,605 @@
...
+}
+
+// Send a request from the host to the SP, returning the number of bytes sent.
+async fn send_outgoing_request(
+    socket: &UdpSocket,
+    peer: &SocketAddrV6,
+    request: &mut OutstandingHostRequest,
+    buf: &mut [u8],
+) -> Result<usize, Error> {
+    let mut msg_size = match hubpack::serialize(buf, &request.request.message) {
```
mkeeter, line 575: We talked about this in matrix, but it seems reasonable to unwrap this.

controller/src/lib.rs
```diff
@@ -2,10 +2,605 @@
...
+    /// requests over the network are received, they'll be passed into the
+    /// handler, and the yielded response forwarded back to the SP.
+    pub async fn new<H, F>(log: Logger, request_handler: H) -> Result<Self, Error>
+    where
+        H: Fn(SpRpcRequest) -> F + Send + Sync + 'static,
+        F: Future<Output = Result<HostRpcResponse, Error>> + Send,
+    {
+        // TODO-correctness We probably want to accept a specific address as
+        // part of the construction of this object.
+        let local_addr = SocketAddrV6::new(Ipv6Addr::UNSPECIFIED, PORT, 0, 0);
```
jgallagher, line 142: May not be needed on this PR, but we will almost certainly need to set the `scope_id` here and in `join_multicast_v6()` below based on the interface that this socket should use. The way MGS does this currently (which may not be right long term) is to wait for the interface to exist, determined by calling `if_nametoindex()` in a loop until it succeeds: https://github.com/oxidecomputer/management-gateway-service/blob/d7791e4d02571b8414b4c2976b0439234aa4d8c7/gateway-sp-comms/src/single_sp/startup.rs#L219-L246

controller/src/lib.rs
```diff
@@ -2,10 +2,605 @@
...
+    log: Logger,
+    mut incoming_request_rx: mpsc::Receiver<SpRpcRequest>,
+    outgoing_response_tx: mpsc::Sender<HostRpcResponse>,
+    request_handler: H,
+) where
+    H: Fn(SpRpcRequest) -> F + Send + Sync + 'static,
+    F: Future<Output = Result<HostRpcResponse, Error>> + Send,
+{
+    loop {
+        let incoming_request = incoming_request_rx.recv().await.unwrap();
```
mkeeter, line 599: Is it possible for the sender to close the channel here?

If so, we could get a clean exit for free with something like
```rust
while let Some(incoming_request) = incoming_request_rx.recv().await {
    // ...
```

messages/src/lib.rs
```diff
@@ -16,8 +16,9 @@ use hubpack::SerializedSize;
 use serde::Deserialize;
 use serde::Serialize;
 
-/// The maximum size of the body of a message.
-pub const MAX_MESSAGE_SIZE: usize = 1024;
+/// The maximum size of the payload of a message, i.e., any trailing data after
+/// the [`message::Message`] contents.
+pub const MAX_PAYLOAD_SIZE: usize = 1024;
```
mkeeter, line 21: Thanks for the renaming here 😄

controller/src/lib.rs
```diff
@@ -2,10 +2,605 @@
...
+async fn io_loop(
+    log: Logger,
+    socket: UdpSocket,
+    mut outgoing_request_rx: mpsc::Receiver<OutstandingHostRequest>,
+    mut outgoing_response_rx: mpsc::Receiver<HostRpcResponse>,
+    incoming_request_tx: mpsc::Sender<SpRpcRequest>,
+) {
+    let mut recv_buf = [0u8; MAX_PAYLOAD_SIZE + Message::MAX_SIZE];
+    let mut send_buf = recv_buf.clone();
+    let peer_addr = SocketAddrV6::new(Ipv6Addr::from(ADDR), PORT, 0, 0);
```
jgallagher, line 311: Will need to set the `scope_id` here too.

messages/src/lib.rs
```diff
@@ -143,6 +161,11 @@ impl PortMask {
             Ok(Self(1 << index))
         }
     }
+
+    /// Return the number of transceivers addressed by `self.
+    pub const fn n_transceivers(&self) -> usize {
```
mkeeter, line 166: Nit: is this name standard / pulled from somewhere else?

Otherwise, I'd vote for something obvious, even if it's a bit longer – maybe `selected_transceiver_count`?

controller/src/lib.rs
```diff
@@ -2,10 +2,605 @@
...
+                }
+            }
+
+            // If we _do_ have an outstanding request, we need to resend it
+            // periodically until we get a response. Wait for up to the resend
+            // interval of inactivity, and then possibly retry.
+            _ = retry_timeout, if outstanding_request.is_some() => {
+                debug!(log, "timed out without response, retrying");
+                let request = outstanding_request.as_mut().unwrap();
+                send_outgoing_request(&socket, &peer_addr, request, &mut send_buf).await.unwrap();
```
jgallagher, line 377: Is this `.unwrap()` safe? If we previously failed to send up on line 367, we could end up in this branch and presumably fail to send again, right?

controller/src/lib.rs
jgallagher: Bit of clever syntax I think I picked up from @mkeeter for cases like this:

```rust
buf[msg_size..][..data.len()].copy_from_slice(data);
```

to avoid having to remember to do the addition.

jgallagher: Do we need to check and return an error if `data` doesn't run past the end of `buf`? Currently this panics, which is maybe fine if we know buf is always big enough.

controller/src/lib.rs
bnaecker: Good point, thank you. I'll move this until we've guaranteed we're looking at a request from the SP.

Also, FYI -- These argument signatures are getting a bit long, especially since I'm going to add some DTrace probes. I'm planning to transform the `io_loop` function into a POD type and make `send_version_mismatch` etc inherent methods. Apologies for the churn on this!

controller/src/lib.rs
bnaecker: Oh, nice idea. I don't know yet if we'll ever see `None` here, since it depends on how I end up using this type in Dendrite. I.e., if we drop the sender before we abort this task. But I think your suggestion will be cleaner regardless, thanks!

controller/src/lib.rs
bnaecker: That's a good question. I've not really written the part of the `Controller` that writes data, which I think is the only one accepting data on the host side at this point. But I was planning to have a function `Controller::write()`, which accepts `data` and returns an error if the data is too big to ever be serialized. I'm not sure how that'll play out, but that would mean that we should never overrun the buffer.

messages/src/lib.rs
bnaecker: It's totally arbitrary! I'll rename to your suggestion, much clearer thank you.

controller/src/lib.rs
bnaecker: Agreed, and I'm also hoping to ignore this right now. (Though I should make an issue to track.)

Just thinking out loud for a minute: thi type will be consumed by `dpd`, the main Dendrite binary. That will run inside the switch zone, launched by the sled agent (bootstrap agent, really). All of the zones the sled agent currently kicks off have a VNIC over the inter-zone etherstub, along with an IPv6 link-local address. The interface name could be provided to `dpd`, just like it is for MGS. So `if_nametoindex` seems like the right answer here.

Something about the retry loop seems odd, since if it doesn't exist, then the sled agent didn't start things up correctly. But as you pointed out, that's a longer term concern! Thanks.

controller/src/lib.rs
bnaecker: No, it's not. Thanks for flagging. I'll use the same pattern as above, logging and continuing the loop.

controller/src/lib.rs
bnaecker: I'm actually not sure I understand. If we can't deserialize the message, the only thing we can check is the version. We don't know what kind of message it was supposed to be, so we can't use that to decide how to respond. We can still send a protocol error, but we can't do something like decide whether this was a request.

So maybe a hierarchy like this would work:

- If the message fails to deserialize, either drop the message or send a protocol error (or make a specific error for this).
- If the message serializes but is of the wrong type, send a protocol error.
- If the message is of the right type but wrong version, send a version mismatch error.

mkeeter: That's a good point – I forgot that we can't decode the message _type_ from the header alone.

I still think we should try decoding the header, so that we can **log** a version mismatch, but think that we should drop packets with version mismatches (regardless of whether deserialization succeeds or not): if the version is mismatched, then even if deserialization "succeeds", we can't necessarily trust the data.

(also, if we replied, I could see two devices with different versions getting into a loop of spamming version mismatch errors at each other)

bnaecker: > That's a good point – I forgot that we can't decode the message type from the header alone.

Another option would be to include the type as a separate enum in the header, but that feels a bit duplicative.

> I still think we should try decoding the header, so that we can log a version mismatch, but think that we should drop packets with version mismatches (regardless of whether deserialization succeeds or not): if the version is mismatched, then even if deserialization "succeeds", we can't necessarily trust the data.

Good point. That seems like a recipe for deserializing the _wrong_ thing, rather than just failing to deserialize.

To summarize, the plan is to:

- If deserializing the message fails, deserialize the header. Log the failure and then drop the packet.
- If deserializing the message succeeds, but there is a version mismatch, log and drop the packet. This is regardless of the _type_ of message, on the theory that we may not be able to accurately tell whether the message violates the protocol. Do _not_ send a version mismatch error.
- If deserializing the message succeeds _and_ the version is fine, check that it satisfies the protocol. If not, send a protocol error.

> (also, if we replied, I could see two devices with different versions getting into a loop of spamming version mismatch errors at each other)

As hilarious as that would be, we should probably avoid it :)

controller/src/lib.rs
bnaecker: This is obviated by the other comment thread, pointing out that if the version is wrong we don't want to trust anything about the message. We're opting to just drop it now, and never send version mismatches over the wire.

controller/src/lib.rs
bnaecker: See https://github.com/oxidecomputer/transceiver-control/issues/14

controller/src/lib.rs
mkeeter: 👍🏻

bnaecker:
Ok, pretty sure I've addressed the substantive changes. As I mentioned, I reorganized a bit, pushing all the data for the IO loop into a single type. That shouldn't be too big a functional change, though the straight diff might look big. Let me know if there are other questions / issues, and I'll start testing on hardware in the meantime!

controller/src/lib.rs
bnaecker: I don't know, but it is a clear violation of the protocol. OTOH, it's true that the SP can't do anything except log it, without an update to its software. I'm ambivalent, I guess. One issue I see is, if we don't send a protocol error in this case, why would we ever send one?

controller/src/lib.rs
bnaecker: Done in 3e4908a9ef71d40c64b75553cccc4d9e0d0344cd

controller/src/lib.rs
mkeeter: ...Still seems the same as above?  I don't think `io_task` and `request_task` are doing exactly the same thing.

controller/src/lib.rs
mkeeter: 👍🏻

controller/src/lib.rs
bnaecker: D'oh, got lost in the local changes. Thanks, will fix now.

bnaecker: Fixed in 95ab907

controller/src/bin/mock-sp.rs
```diff
@@ -0,0 +1,55 @@
...
+use std::net::SocketAddrV6;
+use tokio::net::UdpSocket;
+use transceiver_messages::message::HostRequest;
+use transceiver_messages::message::Message;
+use transceiver_messages::message::MessageBody;
+use transceiver_messages::message::SpResponse;
+use transceiver_messages::MAX_PAYLOAD_SIZE;
+use transceiver_messages::PORT;
+
+const PEER: Ipv6Addr = Ipv6Addr::new(0xff02, 0, 0, 0, 0, 0, 0x1de, 2);
```
mkeeter, line 13: Hmmm, I'd like to use `Ipv6Addr::from(transceiver_messages::ADDR)`, but that's not `const`.

It might be worth it to de-`const` this (moving it into `main`), so that we don't have to specify the IP address in two places?

mkeeter (approved):

controller/src/bin/mock-sp.rs
bnaecker: Sure, sounds fine, change incoming.

bnaecker: Done in b285b3d

PR #16, opened 2022-11-18
Support for some configuration values

bnaecker:
- Adds a `Config` type for specifying the configuration for a `Controller`. Currently provides knobs for retry duration/interval, local/remote IP addresses, and the IP interface. Plumbs these up to the `xcvradm` CLI tool.
- Adds a few convenience constructors to `PortMask` and `ModuleId`.
- Fixes bug where we send back a ProtocolError if the SP responds incorrectly to our outstanding request, but then just continues to try to send it. That leads to a silly ping-pong of error messages back and forth. Now we fail the request if we ever get back a ProtocolError, since it could never succeed.

### DIFF
--- a/controller/src/bin/xcvradm.rs
+++ b/controller/src/bin/xcvradm.rs
@@ -1,8 +1,15 @@
+use clap::Parser;
+use clap::Subcommand;
 use slog::Drain;
+use slog::Level;
+use std::net::Ipv6Addr;
+use std::time::Duration;
+use transceiver_controller::Config;
 use transceiver_controller::Controller;
 use transceiver_controller::Error;
 use transceiver_controller::HostRpcResponse;
 use transceiver_controller::SpRpcRequest;
+use transceiver_messages::message::Status;
 use transceiver_messages::mgmt::sff8636;
 use transceiver_messages::mgmt::MemoryRead;
 use transceiver_messages::Error as MessageError;
@@ -13,11 +20,99 @@ async fn request_handler(_: SpRpcRequest) -> Result<HostRpcResponse, Error> {
     Err(Error::Protocol(MessageError::ProtocolError))
 }
 
+fn parse_xvcr_list(s: &str) -> Result<Vec<u8>, String> {
+    s.split(",")
+        .map(|x| {
+            x.parse()
+                .map_err(|_| String::from("invalid transceiver index"))
+        })
+        .collect()
+}
+
+fn parse_log_level(s: &str) -> Result<Level, String> {
+    s.parse().map_err(|_| String::from("invalid log level"))
+}
+
+#[derive(Parser)]
+struct Args {
+    #[command(subcommand)]
+    cmd: Cmd,
+
+    /// The FPGA whose transceivers to address.
+    #[arg(short, long, default_value_t = 0)]
+    fpga_id: u8,
+
+    /// The comma-separated list of transcievers on the FPGA to address.
+    ///
+    /// The default is all transceivers on an FPGA.
+    #[arg(short, long)]
+    transceivers: Option<String>,
+
+    /// The source IP address on which to listen for messages.
+    #[arg(short, long, default_value_t = Ipv6Addr::UNSPECIFIED)]
+    address: Ipv6Addr,
+
+    /// The source interface on which to listen for messages.
+    #[arg(short, long)]
+    interface: String,
+
+    /// The unicast peer address to assume.
+    ///
+    /// The protocol is normally run using a multicast address, but a single
+    /// peer may optionally be specified.
+    #[arg(short, long)]
+    peer: Option<Ipv6Addr>,
+
+    /// The maximum number of retries before failing a request.
+    #[arg(short, long)]
+    n_retries: Option<usize>,
+
+    /// The retry interval for requests, in milliseconds.
+    #[arg(
+        short,
+        long,
+        default_value_t = 1000,
+        value_parser = clap::value_parser!(u64).range(1..=10000)
+    )]
+    retry_interval: u64,
+
+    /// The log-level.
+    #[arg(
+        short,
+        long,
+        default_value_t = Level::Info,
+        value_parser = parse_log_level
+    )]
+    log_level: Level,
+}
+
+#[derive(Subcommand)]
+enum Cmd {
+    /// Return the status of the addressed modules, such as presence, power
+    /// enable, and power mode.
+    Status,
+
+    /// Read the lower page of a set of transceiver modules.
+    ReadLower { offset: u8, len: u8 },
+}
+
 #[tokio::main]
 async fn main() {
+    let args = Args::parse();
+    let config = Config {
+        address: args.address,
+        interface: args.interface,
+        peer: args
+            .peer
+            .unwrap_or(Ipv6Addr::from(transceiver_messages::ADDR)),
+        n_retries: args.n_retries,
+        retry_interval: Duration::from_millis(args.retry_interval),
+    };
+
     let decorator = slog_term::TermDecorator::new().build();
     let drain = slog_term::FullFormat::new(decorator).build().fuse();
     let drain = slog_async::Async::new(drain).build().fuse();
+    let drain = slog::LevelFilter::new(drain, args.log_level).fuse();
     let log = slog::Logger::root(drain, slog::o!());
 
     let request_handler_log = log.new(slog::o!("name" => "request_handler"));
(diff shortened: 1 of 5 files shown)


controller/src/lib.rs
```diff
@@ -254,13 +335,23 @@ impl Controller {
...
+
+        // If we get back a ReadFailed error, one possibility is that we asked
+        // to read a transceiver that's not present.
+        let data = match reply.message.body {
+            MessageBody::SpResponse(SpResponse::Error(MessageError::ReadFailed)) => {
+                return Err(Error::ReadFailed);
+            }
+            MessageBody::SpResponse(SpResponse::Error(e)) => return Err(Error::from(e)),
+            MessageBody::SpResponse(SpResponse::Read(_)) => reply.data.unwrap(),
+            _ => panic!("Unexpected message type"),
```
jgallagher, line 348: This might be okay, but it's not obvious that it is - does `self.rpc()` guarantee that `reply.message.body` will fall into one of the three cases above? Could this return something like `Error::UnexpectedMessageType(maybe_some_details_maybe_not)` instead of panicking?

controller/src/lib.rs
```diff
@@ -254,13 +335,23 @@ impl Controller {
...
         // We expect data to be a flattened vec of vecs, with the data from each
         // referenced transceiver. Split it into chunks sized by the number of
         // bytes we expected to read.
-        let data = reply
-            .data
-            .unwrap()
+        let data = data
             .chunks_exact(usize::from(read.len()))
             .map(Vec::from)
             .collect::<Vec<_>>();
```
jgallagher, line 357: There's an assert just below this line in the diff and not new in the PR, but I have a similar question as on the panic above: does `self.rpc()` guarantee that `data.len()` is exactly what we expect in this assertion?

controller/src/lib.rs
```diff
@@ -468,15 +562,27 @@ impl IoLoop {
...
                 // If we _do_ have an outstanding request, we need to resend it
                 // periodically until we get a response. Wait for up to the resend
                 // interval of inactivity, and then possibly retry.
                 _ = self.resend.tick(), if self.outstanding_request.is_some() => {
-                    debug!(self.log, "timed out without response, retrying");
-                    self.send_outgoing_request(&mut tx_buf).await;
-                    self.resend.reset();
+                    if n_retries < self.n_retries {
+                        debug!(self.log, "timed out without response, retrying");
+                        self.send_outgoing_request(&mut tx_buf).await;
```
jgallagher, line 574: Nit - Would it make sense for `send_outgoing_request` to call `self.resend.reset()` and/or increment `n_retries` instead of duplicating this between here and line 563-565 above? `n_retries` is currently a local var so that might be awkward (unless it could reasonably be a field on `self`? not sure about that).

controller/src/bin/xcvradm.rs
```diff
@@ -13,29 +20,145 @@ async fn request_handler(_: SpRpcRequest) -> Result<HostRpcResponse, Error> {
     Err(Error::Protocol(MessageError::ProtocolError))
 }
 
+fn parse_xvcr_list(s: &str) -> Result<Vec<u8>, String> {
+    s.split(",")
+        .map(|x| {
+            x.parse()
+                .map_err(|_| String::from("invalid transceiver index"))
+        })
+        .collect()
```
mkeeter, line 29: We could flip the order of the collect / map_err here, which makes it shorter and point-free (but adds a turbofish); I'm not sure if this is an improvement!
```suggestion
    s.split(",")
        .map(str::parse)
        .collect::<Result<_, _>>()
        .map_err(|_| String::from("invalid transceiver index"))
```

controller/src/bin/xcvradm.rs
```diff
@@ -13,29 +20,145 @@ async fn request_handler(_: SpRpcRequest) -> Result<HostRpcResponse, Error> {
...
+    s.parse().map_err(|_| String::from("invalid log level"))
+}
+
+#[derive(Parser)]
+struct Args {
+    #[command(subcommand)]
+    cmd: Cmd,
+
+    /// The FPGA whose transceivers to address.
+    #[arg(short, long, default_value_t = 0)]
```
mkeeter, line 42: TIL about `default_value_t` versus `default_value` 🤯

controller/src/bin/xcvradm.rs
```diff
@@ -13,29 +20,145 @@ async fn request_handler(_: SpRpcRequest) -> Result<HostRpcResponse, Error> {
...
+    fpga_id: u8,
+
+    /// The comma-separated list of transcievers on the FPGA to address.
+    ///
+    /// The default is all transceivers on an FPGA.
+    #[arg(short, long)]
+    transceivers: Option<String>,
```
mkeeter, line 49: On second thought, I think we can do the parsing here (instead of in `parse_xvcr_list`):
```suggestion
    #[clap(long, short, use_value_delimiter = true)]
    transceivers: Option<Vec<u8>>,
```

controller/src/bin/xcvradm.rs
```diff
@@ -13,29 +20,145 @@ async fn request_handler(_: SpRpcRequest) -> Result<HostRpcResponse, Error> {
...
+
 #[tokio::main]
 async fn main() {
+    let args = Args::parse();
+    let config = Config {
+        address: args.address,
+        interface: args.interface,
+        peer: args
+            .peer
+            .unwrap_or(Ipv6Addr::from(transceiver_messages::ADDR)),
```
mkeeter, line 107:
```suggestion
            .unwrap_or_else(|| Ipv6Addr::from(transceiver_messages::ADDR)),
```

controller/src/bin/xcvradm.rs
```diff
@@ -13,29 +20,145 @@ async fn request_handler(_: SpRpcRequest) -> Result<HostRpcResponse, Error> {
...
+            let data = controller.read(modules, read).await.unwrap();
+            print_read_data(modules, data);
+        }
+    }
+}
+
+fn print_module_status(modules: ModuleId, status: Vec<Status>) {
+    println!("FPGA\tPort\tStatus");
+    for (port, status) in modules.ports.to_indices().zip(status.into_iter()) {
+        println!("{}\t{port}\t{status:?}", modules.fpga_id);
```
mkeeter, line 150: In Humility tables, we use fixed-width format strings (rather than tabs, which can vary from terminal to terminal):
```suggestion
        println!("{:4>} {port:4>} {status:?}", modules.fpga_id);
```
(this comment is generally applicable to all of the `print_*` functions here)

controller/src/lib.rs
```diff
@@ -232,8 +308,13 @@ impl Controller {
...
         };
-        let _reply = self.rpc(request).await.unwrap();
-        todo!();
+        let reply = self.rpc(request).await?;
+        Ok(reply
+            .data
+            .unwrap()
+            .into_iter()
+            .map(|x| Status::from_bits(x).unwrap())
+            .collect())
```
mkeeter, line 317: Use `MissingData` here?

e.g.
```suggestion
        let reply = self.rpc(request).await?;
        let data = reply.data.ok_or(MessageError::MissingData)?;
        Ok(data
            .into_iter()
            .map(|x| Status::from_bits(x).unwrap())
            .collect())
```

mkeeter:
I added a few small comments; you may also want to run stuff through `clippy` and see what it thinks.

controller/src/lib.rs
bnaecker: I think I tried that but found it awkward since `send_outgoing_request` takes a shared rather than exclusive reference. But there's no reason it can't be modified to take `&mut self`! I'll add the number of retries to the type as well, and do those modifications in that method. Thanks!

controller/src/lib.rs
bnaecker: Yes, `self.rpc()` guarantees that the length of the data is exactly what we expect. That's done using the `Message::expected_data_len` method. So this implicit panic was intentional.

Now that I think about it, one fallout of https://github.com/oxidecomputer/transceiver-control/issues/17 might be that this would need to change slightly. Short reads may become possible, if a module is unplugged while the request is being processed by the SP. It shouldn't change this part here, but the check for the expected data length should use the _response_ message to determine that, rather than the request.

controller/src/lib.rs
bnaecker: I think I meant to write `todo!()` here rather than `panic!()`! Still, I think we could return an error here. `self.rpc()` guarantees that we get a `MessageBody::SpResponse(_)`, but not what that inner variant is. I should return a protocol error or similar, you're right. Thanks!

controller/src/lib.rs
bnaecker: Actually, this already uses the response to determine how much data to expect. I think this is OK at this point.

controller/src/bin/xcvradm.rs
bnaecker: TIL about `use_value_delimiter`! 😆

controller/src/bin/xcvradm.rs
bnaecker: Nice, that's a good idea thanks!

controller/src/lib.rs
bnaecker: I think we could, but `self.rpc` should already check that the length of `data` is exactly what we expect. That's done [here](https://github.com/oxidecomputer/transceiver-control/blob/939b4b2d89571ef1ec0bef25bc8fb2ec8eaa4fd0/controller/src/lib.rs#L729-L755). Does that suffice?

controller/src/bin/xcvradm.rs
bnaecker: That looks a bit nicer to me, thanks!

bnaecker:
At @mkeeter's suggestion, I ran Clippy against this. It has a few benign suggestions, e.g., also deriving `Eq` for types where that's possible. Nothing is substantial, surprisingly!

controller/src/lib.rs
mkeeter: Aha, that was far enough away that I forgot about it.  Maybe use something like `.expect("checked earlier")` as a reminder?

mkeeter: (although per [parse-don't-validate](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/), this suggests we should have a different type for "message with known-correct `data` member" 😃)

mkeeter (approved):

jgallagher (approved):

controller/src/lib.rs
bnaecker: Yeah, that's fair! The `SpResponse` type really _is_ the type representing that. It just needs a fallible constructor or something, so that we can't build one that _should_ carry data but does not.

PR #18, opened 2022-11-21
Add hardware error type and bump to V2

mkeeter:
This adds a fine-grained `HwError` which provides details on SP -> FPGA -> transceiver errors.

It also introduces a few new member to the top-level `Error` enum.

### DIFF
--- a/controller/src/lib.rs
+++ b/controller/src/lib.rs
@@ -297,7 +297,7 @@ impl Controller {
     // Return a header using the next available message ID.
     fn next_header(&self) -> Header {
         Header {
-            version: message::version::V1,
+            version: message::version::V2,
             message_id: self.message_id.fetch_add(1, Ordering::SeqCst),
         }
     }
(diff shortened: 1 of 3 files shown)


controller/src/lib.rs
```diff
@@ -297,7 +297,7 @@ impl Controller {
     // Return a header using the next available message ID.
     fn next_header(&self) -> Header {
         Header {
-            version: message::version::V1,
+            version: message::version::V2,
```
bnaecker, line 300: On thing that occurs to me is we might want to add a `message::version::CURRENT` const "alias", so that this library is always compiled against the latest version in the source tree. Thoughts?

messages/src/lib.rs
```diff
@@ -89,6 +101,80 @@ impl std::fmt::Display for Error {
...
+    ReadSetupFailed,
+
+    /// Reading back the read buffer failed
+    ReadBufFailed,
+
+    /// The write setup operation failed
+    WriteSetupFailed,
+
+    /// The write operation failed
+    WriteFailed,
```
bnaecker, line 168: Is this a catch-all for a generic write failure? Or does this mean something specific in terms of the SP <-> FPGA interface?

bnaecker (approved):
Looks good! A couple questions, but not block. Thanks!

controller/src/lib.rs
mkeeter: Yeah, that makes sense to me – done.

Issue #20, opened 2022-11-22
Want library API and CLI command for resetting modules


bnaecker:

bnaecker:
Completed in #25

PR #25, opened 2022-11-22
Add a reset method

bnaecker:
- Adds the `Controller::reset` method for reseting some transceivers.
- Adds a `xcvradm reset` subcommand for consuming it.


bnaecker:
Addresses #20

PR #26, opened 2022-11-23
WIP: Adds support for writing to module memory

bnaecker:
- Reworks `xcvradm` command to require the management interface it expects modules to use when reading / writing.
- Checks the management interface implied by the `Memory{Read,Write}` objects in the `Controller::{read,write}` methods.
- Adds errors for indicating that a module doesn't match the expected interface.
- Adds error variant indicating that a read / write is too big for the provided memory page based on its management interface. This is split off from the previous generic "invalid memory access" error, which didn't distinguish between something being out of bounds and something being in-bounds, but too large for a module (e.g., 4-byte writes for SFF-8636).
- Adds `xcvradm {read,write}-{lower,upper}` commands.
- Adds option to print read data in binary.

PR #30, opened 2022-11-30
Check RPC response matches expected message type

bnaecker:


bnaecker:
Straggling commit that should have been part of #26. This fixes a few more places with the issue @Aaron-Hartwig found, where we didn't check the actual response of an RPC matches the message type we expect.

Aaron-Hartwig (approved):

Issue #32, opened 2022-12-07
Implement request handler via message passing


bnaecker:
Right now, we have one `Controller` type in the `transceiver_controller` crate. Users can configure which interface and IP address to use for sending requests, but not the port. They're also required to specify a request-handler at the moment, basically an async function for responding to SP requests.

The first of these makes it difficult to create more than one host-side client. Without the ability to set the port, we are required to use a different IP address or interface. Both require creating something the only purpose of which is to send these messages. We could use the port instead.

The latter of these makes _sharing_ a controller difficult. The request handler is mostly useful when it has some kind of state attached to it, which can be done by implementing a trait on some type for the handler. But that can lead to reference cycles if, for example, we want to store the controller in the type that implements that trait itself. So, if sharing a controller is difficult, we'd like to have multiple independent ones if possible. That's difficult because of the above, so the situation could be improved.

The motivating use case here is integrating this into the BF SDE. The Sidecar platform implementation needs a controller to actually operate on the transceivers, but it doesn't currently need to have a request handler. Furthermore, even if it _did_ need a request handler, it's not possible to have two request handlers for the same request. That's for a few reasons -- first, there are obvious dueling administrator-style problems there; and second, who would receive the actual UDP packet? The Hubris task sends an unsolicited request to one address, and exactly zero or one clients can be bound to that address.

All of this suggests a redesign of the crate.

### Unidirectional client

The first proposal is to create a new type, say `Client`, which only allows making requests _to_ the SP and getting the responses. There is no request handler at all, and the `Client` object binds to a random UDP port assigned by the OS rather than the fixed protocol port. This would let us have both one `Controller` and zero or more `Client`s simultaneously. We might eventually want to restrict the kinds of things this client can do, but that's not super important.

### Handling requests via message-passing

The `Controller` currently uses message passing internally for communicating with the various components. We should use this for the request handler itself. I was initially leery of this for a couple of reasons. First, we need a way to specify a channel on which responses are sent back. But we can use the "embedded oneshot" mechanism we already use internally. After creating a `Controller` users can subscribe to the "SP request channel". The `Controller` puts on that channel the request from the SP and the _oneshot_ channel on which the response should be sent.

[...]

PR #33, opened 2022-12-07
Request handler trait

bnaecker:


bnaecker:
This isn't the long-term direction I want to go (see #32), but it works for now. This adds a trait for request handlers, and implements a dummy one that just logs each request and drops it. It also re-exports a bunch of stuff in the controller crate from the messages crate.

This is needed for compliance, so I'm going to merge this once CI and I are happy.

PR #39, opened 2022-12-08
Don't panic if decoding vendor information fails

bnaecker:


decode/src/lib.rs
```diff
@@ -431,13 +440,20 @@ where
...
+        const BAD: &str = "01";
         let year = std::str::from_utf8(&buf[..2])
-            .unwrap()
+            .unwrap_or(BAD)
             .parse::<i32>()
             .unwrap()
             + 2000;
-        let month: u32 = std::str::from_utf8(&buf[2..4]).unwrap().parse().unwrap();
-        let day: u32 = std::str::from_utf8(&buf[4..6]).unwrap().parse().unwrap();
+        let month: u32 = std::str::from_utf8(&buf[2..4])
```
mkeeter, line 449: Out of curiosity, have we seen this on actual hardware?

bnaecker: Yep. It's cheap OEM DAC parts we have in the lab, though. Nothing that's a "real" vendor part. That's why I had this in there originally as an unwrap.

PR #44, opened 2022-12-15
Handle requests via a channel

bnaecker:
Removes the async-trait for a request handler, and instead opts to take a send-half of a channel at construction. Rather than calling the request handler method directly, the controller now sends the request on that channel along with a oneshot for the reponse. It awaits the response on that oneshot; if it's Some(_), the message is serialized and sent back to the peer over the UDP socket; if it's None, the message is dropped. This approach reduces the complexity around ownership and generally allows more freedom to run the controller and handler in separate tasks.

### DIFF
--- a/controller/Cargo.toml
+++ b/controller/Cargo.toml
@@ -6,7 +6,6 @@ default-run = "xcvradm"
 
 [dependencies]
 anyhow = "1"
-async-trait = "0.1"
 hubpack = "0.1.0"
 serde = "1"
 slog-async = "2"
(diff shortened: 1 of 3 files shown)


bnaecker:
This does one half of #32. We don't have the version of the client that can be built, but does _not_ handle requests yet. That isn't needed right now, but might be if we start running the BF SDE and `xcvradm` on the same host. Those will conflict when trying to bind the same local UDP socket. The "non-handling" version of the client could bind any UDP port, and send to the multicast protocol address, since the SP need not be able to deliver a message to it (either via unicast or multicast).

PR #46, opened 2022-12-22
Don't take an outstanding message unless we get its response

bnaecker:
The `IOLoop` previously took out of the optional outstanding message whenever we got a well-formed response, and _then_ checked if the message ID confirms that it's for the outstanding response. That was premature; we should only remove the outstanding message _after_ verifying the response we have is actually for that message.


jclulow (approved):
Seems reasonable!

Issue #47, opened 2022-12-22
Logging is too verbose


bnaecker:
The logging inside the controller `IOLoop` is too verbose. We print debug messages for basically everything, including sending / receiving packets and messages, and a few other checks. That's been really useful during development, but the logging is too verbose when used in `dpd`, for example. Much of that should be moved to trace-leve logs, or into USDT probes if needed.

bnaecker:
Closed by #48

PR #48, opened 2022-12-22
Reduce logging verbosity

bnaecker:
- Move some of debug logs on the most common code paths to trace-level statements.
- Add more bad-message probes.


bnaecker:
Resolves https://github.com/oxidecomputer/transceiver-control/issues/47

PR #53, opened 2023-01-07
Support a configurable source UDP port

bnaecker:
- Add a `port` to the `Config` and its builder, with a default of the protocol port 11112.
- Add CLI option for selecting this, defaulting to 0 for an arbitrary port.

### DIFF
--- a/controller/src/bin/xcvradm.rs
+++ b/controller/src/bin/xcvradm.rs
@@ -151,6 +151,10 @@ struct Args {
     #[arg(short, long)]
     interface: String,
 
+    /// The source UDP port from which to send messages.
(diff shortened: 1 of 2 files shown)


jclulow:
Is there a reason, in general, for not always using port 0 for this stuff?

bnaecker:
Yeah. The SP will eventually send unsolicited messages to the host, so we need to bind to a known port.

bnaecker:
Friendly ping @jclulow :)

jclulow (approved):
LGTM!  Thanks for this.  It has been working well in compliance so far!

Issue #41, opened 2022-12-08
Host needs to check for software control of `LPMode`


bnaecker:
The BF SDE appears to quickly set byte 93, bit 1 of SFF-8636 modules. That is used by host software to inform the module that the `LPMode` pin should be ignored, and that its functionality is subsumed by byte 93, bit 1. In this case, reading the status back in the host may show `LOW_POWER_MODE` in the status bits, but that's not the whole story. The _actual_ power mode may need to be read from the memory map itself.

This issue tracks putting those smarts into the host, in something like a `get_power_state` command. That looks for software control, and then returns the disposition of the `LPMode` pin or byte 93, bit 2. Note that we may also want to return more power information here, such as the power class.

bnaecker:
Note that this also affects the way we _set_ the power mode! Currently, the `SetPowerMode` command sent from host to SP will tell the SP to assert or de-assert the `LPMode` pin. That's only sufficient if the module isn't currently set to software control of the power state. We need to check for that, then either set LPMode or write the correct value to the module's memory map depending on its kind.

bnaecker:
I'm prototyping this and starting to think it's a bigger change. In particular, I don't believe our current `SetPowerMode` message is a great interface for this. I think we need a message to specifically enable or disable power entirely (the efuse), and also to "pass-through" the `LPMode` signal, the same way we're planning to pass `ResetL`.

The reason is basically that the three `PowerMode`s "work" pretty differently. Turning things off can always be done, regardless of the module. We control that power rail independent of everything else. Turning things _on_, either into low or high-power, works differently. If the module is under hardware control, then we twiddle the `LPMode` pin. If it's under hardware control, we need to write to a specific bit, depending on the kind of module.

That's all fine, but it kind of obviates the purpose of the single `PowerMode` enum in our message. We could still use it in the `Controller` API, but it might be better to have:

- A message for turning on or off the power rails: `SetPowerOn(bool)` as a strawman.
- A message for twiddling `LPMode`: `SetLpActive(bool)` as a strawman.
- We can use the existing memory-write messages for the other case.

All of this is inline with generally keeping the smarts in the host. @Aaron-Hartwig I know you're working on something similar with `ResetL`, does this intersect with that at all? @mkeeter Any thoughts?

To be clear, the existing messaging API will _work_, since we can just not use the enum variants `PowerMode::{Low,High}` if the module is configured for software override of power. It just feels awkward and possibly confusing.

[...]

PR #54, opened 2023-01-25
Add direct control of module power, ResetL, and LpMode

Aaron-Hartwig:
This provides messages to directly control module's power (eFuse enable), reset pin (`ModResetL`), and low-power mode pin (`LpMode`).

This is part of a bundle of commits that need to land at the same time:
FPGA: https://github.com/oxidecomputer/quartz/pull/69
SP: https://github.com/oxidecomputer/hubris/pull/1105


controller/src/lib.rs
```diff
@@ -569,6 +569,72 @@ impl Controller {
...
+    }
+
+    /// Deassert reset for a set of transceiver modules.
+    pub async fn deassert_reset(&self, modules: ModuleId) -> Result<(), Error> {
+        self.no_payload_request(modules, HostRequest::DeassertReset)
+            .await?;
+        Ok(())
+    }
+
+    /// Assert lpmode for a set of transceiver modules.
```
bnaecker, line 600: I might add a note here that this is about the `LPMode` pin itself, and will have no effect if the software override of power control is set.

controller/src/lib.rs
```diff
@@ -569,6 +569,72 @@ impl Controller {
...
+    pub async fn assert_lpmode(&self, modules: ModuleId) -> Result<(), Error> {
+        self.no_payload_request(modules, HostRequest::AssertLpMode)
+            .await?;
+        Ok(())
+    }
+
+    /// Deassert lpmode for a set of transceiver modules.
```
bnaecker, line 607: Same comment here maybe.

messages/src/message.rs
```diff
@@ -108,6 +108,24 @@ pub enum HostRequest {
     /// Request that the modules be reset.
     Reset,
 
+    /// Request that the ResetL line be asserted.
+    AssertReset,
```
bnaecker, line 112: I can't comment on the actual line because GitHub UI is the worst, but we should bump the protocol version number in this change too.

messages/src/message.rs
```diff
@@ -108,6 +108,24 @@ pub enum HostRequest {
...
+    /// Request that the ResetL line be de-asserted.
+    DeassertReset,
+
+    /// Request that the LpMode line be asserted.
+    AssertLpMode,
+
+    /// Request that the LpMode line be de-asserted.
+    DeassertLpMode,
+
+    /// Request that power be enabled to a module should it be inserted.
```
bnaecker, line 123: What happens if there is no module inserted? Does this return an error message?

messages/src/message.rs
```diff
@@ -108,6 +108,24 @@ pub enum HostRequest {
...
+    EnablePower,
+
+    /// Request that power be disabled to a module should it be inserted.
+    DisablePower,
+
```
bnaecker, line 128: I also can't comment on the relevant line, but I think we'd want to remove the `SetPowerMode` variant entirely from the protocol. The host-side and controller can expose that idea, and then control either the efuse, LPMode pin, or the correct bytes of the module memory map depending on which mode is requested and whether the module responds to LPMode.

I think it's enough to remove this variant; move the `PowerMode` enum to `controller/src/lib.rs`, and then maybe replace the body of the `Controller::set_power()` function with a `todo!()`. The implementation of that is going to change significantly to use these new messages, which seems better in a separate PR.

messages/src/lib.rs
```diff
@@ -187,25 +187,50 @@ pub enum HwError {
...
 
-    /// Failed to read the `EN` register
-    #[cfg_attr(any(test, feature = "std"), error("Failed to read `EN` register"))]
-    EnableReadFailed,
+    /// Failed while adjusting the power mode
+    #[cfg_attr(
+        any(test, feature = "std"),
+        error("Failed while adjusting the power mode")
+    )]
+    PowerModeChangeFailed,
```
bnaecker, line 195: I was confused by this, specifically how it's different from `PowerEnableWriteFailed`. But I looked at the related PR on the Hubris side, and I think I get it.

Just checking my understanding. When the `SetPowerMode` message is sent, the SP does a sequence of a few actions. For example, if the mode is `PowerMode::Low`, it seems to: enable the efuse; assert `LPMode`; and de-assert `ResetL`.

If that's correct, then I think we should remove that entire message and related code in Hubris. The main reason is that we're giving the host direct control over the `LPMode` and `ResetL` pins specifically because we don't want Hubris to have to worry about that. For example, that whole operation is kind of lying if the module has software override of power control. The request to Hubris will succeed, but the module may not change its power mode at all.

It seems we'll instead want to put that kind of operation in the host `Controller`. The alternative is putting the ability to check for software override of power control into Hubris, which I don't believe we're interested in.

bnaecker (commented):
Thanks for taking this on  @Aaron-Hartwig! I think we actually want to go a bit further, and not include the `SetPowerMode` in the messaging protocol at all. See the comments on specific lines for the rationale, and of course let me know if there are any questions or I'm missing something!

messages/src/message.rs
Aaron-Hartwig: It does not. The `should it be inserted` is loadbearing here as this message does not explicitly turn on/off the eFuse, it actually just tells the FPGA _if it should_ turn on/off the eFuse when it sees a module inserted. So if there is no module, the power will be disabled regardless if the bit is set.

messages/src/message.rs
Aaron-Hartwig: I'm cool with removing `SetPowerMode`. It felt like a temporary, parallel path anyway.

messages/src/lib.rs
Aaron-Hartwig: Your understanding is correct. I had both methods of control there since I knew in one sense the host wanted control of the individual signals and in another it may not. I am a fan of handling that abstraction at a layer higher than this!

bnaecker: Sweet! I think let's snip all that out now, so we don't have the vestigial and hard-to-use `SetPowerMode` message at all. I can follow up implementing the new `Controller::set_power` function described in https://github.com/oxidecomputer/transceiver-control/issues/41.

bnaecker (approved):
Excellent!

PR #55, opened 2023-01-30
Handles software control of power mode when setting module power

bnaecker:
This updates the various subcommands for managing a module's power state to handle the case where software override of power control is set. It keeps in place the existing, direct-control subcommands for setting the disposition of hardware signals, and changes the `power_mode` and `set_power_mode` functions to check for software override when operating.

Note that one goal here is always keep the hardware and software signals in sync. That is, if we request low-power mode, we set both the hardware `LPMode` pin and the right bit in the module memory map. The idea is that if one toggles the software-override, the power mode itself will not change.

### DIFF
--- a/controller/src/bin/xcvradm.rs
+++ b/controller/src/bin/xcvradm.rs
@@ -63,8 +63,8 @@ fn parse_transceivers(s: &str) -> Result<Transceivers, String> {
         "all" => Ok(Transceivers::All),
         "present" => Ok(Transceivers::Present),
         "off" => Ok(Transceivers::PowerMode(PowerMode::Off)),
-        "low-power" => Ok(Transceivers::PowerMode(PowerMode::Low)),
-        "hi-power" => Ok(Transceivers::PowerMode(PowerMode::High)),
+        "low-power" | "lp" => Ok(Transceivers::PowerMode(PowerMode::Low)),
+        "hi-power" | "high-power" | "hp" => Ok(Transceivers::PowerMode(PowerMode::High)),
         "sff" => Ok(Transceivers::Kind(ManagementInterface::Sff8636)),
         "cmis" => Ok(Transceivers::Kind(ManagementInterface::Cmis)),
         _maybe_list => {
@@ -211,10 +211,24 @@ enum Cmd {
         mode: PowerMode,
     },
 
-    /// Enable the hot swap controller for the addressed modules
+    /// Return the power mode of the addressed modules.
+    ///
+    /// This takes into account whether a module has specified software override
+    /// of power control.
+    Power,
+
+    /// Enable the hot swap controller for the addressed modules.
+    ///
+    /// Note that this is a lower-level method for specifically controlling the
+    /// hot swap controller directly. See the `set-power` subcommand for a
+    /// higher-level interface to set the module power to a specific state.
     EnablePower,
 
-    /// Disable the hot swap controller for the addressed modules
+    /// Disable the hot swap controller for the addressed modules.
+    ///
+    /// Note that this is a lower-level method for specifically controlling the
+    /// hot swap controller directly. See the `set-power` subcommand for a
+    /// higher-level interface to set the module power to a specific state.
     DisablePower,
 
     /// Assert ResetL for the addressed modules.
@@ -224,9 +238,17 @@ enum Cmd {
     DeassertReset,
 
     /// Assert LpMode for the addressed modules.
+    ///
+    /// Note that this is a lower-level method for specifically controlling the
+    /// LPMode hardware signal directly. See the `set-power` subcommand for a
+    /// higher-level interface to set the module power to a specific state.
     AssertLpMode,
 
     /// Deassert LpMode for the addressed modules.
+    ///
+    /// Note that this is a lower-level method for specifically controlling the
+    /// LPMode hardware signal directly. See the `set-power` subcommand for a
+    /// higher-level interface to set the module power to a specific state.
     DeassertLpMode,
 
     /// Read the SFF-8024 identifier for a set of modules.
@@ -483,6 +505,14 @@ async fn main() -> anyhow::Result<()> {
                 .context("Failed to set power mode")?;
         }
 
+        Cmd::Power => {
+            let states = controller
+                .power_mode(modules)
+                .await
+                .context("Failed to get power mode")?;
+            print_power_mode(modules, states);
+        }
+
         Cmd::EnablePower => {
             controller
                 .enable_power(modules)
@@ -697,30 +727,23 @@ async fn address_transceivers(
         Transceivers::PowerMode(mode) => {
             // Fetch all power modes, and find those which match.
             let modules = ModuleId::all_transceivers(fpga_id);
-            let status = controller
-                .status(modules)
+            let module_modes = controller
+                .power_mode(modules)
                 .await
-                .context("Failed to retrieve module status")?;
-            let predicate = match mode {
-                PowerMode::Off => |st: Status| !st.contains(Status::POWER_GOOD),
-                PowerMode::Low => {
-                    |st: Status| st.contains(Status::POWER_GOOD | Status::LOW_POWER_MODE)
-                }
-                PowerMode::High => |st: Status| {
-                    st.contains(Status::POWER_GOOD) && !st.contains(Status::LOW_POWER_MODE)
-                },
-            };
-            filter_ports(modules.ports, status, predicate)
+                .context("Failed to retrieve module power mode")?;
+            filter_ports(modules.ports, module_modes, |m| m.0 == mode)
         }
         Transceivers::Kind(kind) => {
-            // Fetch all modules that have power enabled, thus are readable.
+            // Fetch all modules that are in at least low-power mode, and thus
+            // readable.
             let modules = ModuleId::all_transceivers(fpga_id);
-            let status = controller
-                .status(modules)
+            let power_modes = controller
+                .power_mode(modules)
                 .await
-                .context("Failed to retrieve module status")?;
-            let readable =
-                filter_ports(modules.ports, status, |st| st.contains(Status::POWER_GOOD));
+                .context("Failed to retrieve module power mode")?;
+            let readable = filter_ports(modules.ports, power_modes, |m| {
+                matches!(m.0, PowerMode::Low | PowerMode::High)
+            });
 
             // Then the management interface for those.
             let modules = ModuleId {
@@ -747,6 +770,23 @@ async fn address_transceivers(
 
 // Column width for printing data below.
 const WIDTH: usize = 4;
+const POWER_WIDTH: usize = 5;
+
+fn print_power_mode(modules: ModuleId, modes: Vec<(PowerMode, Option<bool>)>) {
+    println!("FPGA Port Power Software override");
+    for (port, (mode, override_)) in modules.ports.to_indices().zip(modes.into_iter()) {
+        let over = match override_ {
+            None => "-",
+            Some(true) => "Yes",
+            Some(false) => "No",
+        };
+        let mode = format!("{mode:?}");
+        println!(
+            "{:>WIDTH$} {port:>WIDTH$} {mode:POWER_WIDTH$} {over}",
+            modules.fpga_id
+        );
+    }
+}
 
 fn print_module_status(modules: ModuleId, status: Vec<Status>) {
     println!("FPGA Port Status");
--- a/controller/src/bin/xcvradm.rs
+++ b/controller/src/bin/xcvradm.rs
@@ -2,7 +2,7 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
-// Copyright 2022 Oxide Computer Company
+// Copyright 2023 Oxide Computer Company
 
 //! Command-line tool to administer optical transceivers.
 
@@ -773,7 +773,7 @@ const WIDTH: usize = 4;
 const POWER_WIDTH: usize = 5;
 
 fn print_power_mode(modules: ModuleId, modes: Vec<(PowerMode, Option<bool>)>) {
-    println!("FPGA Port Power Software override");
+    println!("FPGA  Port  Power  Software-override");
     for (port, (mode, override_)) in modules.ports.to_indices().zip(modes.into_iter()) {
         let over = match override_ {
             None => "-",
@@ -782,7 +782,7 @@ fn print_power_mode(modules: ModuleId, modes: Vec<(PowerMode, Option<bool>)>) {
         };
         let mode = format!("{mode:?}");
         println!(
-            "{:>WIDTH$} {port:>WIDTH$} {mode:POWER_WIDTH$} {over}",
+            "{:>WIDTH$}  {port:>WIDTH$}  {mode:POWER_WIDTH$}  {over}",
             modules.fpga_id
         );
     }
--- a/controller/src/lib.rs
+++ b/controller/src/lib.rs
@@ -2,7 +2,7 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
-// Copyright 2022 Oxide Computer Company
+// Copyright 2023 Oxide Computer Company
 
 //! A host-side control interface to the SP for managing Sidecar transceivers.
 
@@ -114,11 +114,11 @@ pub enum PowerMode {
     /// A module is entirely powered off, using the EFuse.
     Off,
 
-    /// Power is enabled to the module, but the `LPMode` pin is set to high.
+    /// Power is enabled to the module, but module remains in low-power mode.
     ///
-    /// Note: This requires that we never set the `Power_override` bit (SFF-8636
-    /// rev 2.10a, section 6.2.6, byte 93 bit 2), as that defeats the purpose of
-    /// hardware control.
+    /// In this state, modules will not establish a link or transmit traffic,
+    /// but they may be managed and queried for information through their memory
+    /// maps.
     Low,
 
     /// The module is in high-power mode.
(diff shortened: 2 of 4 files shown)


bnaecker:
Some notes here about the new command output. We can use the `xcvradm power` command to always return the real power state, and, if the module is powered, whether the module has set software override of control. Note that if the module is off, we can't read from the map and so can't determine if the software override bits are set:

```bash
ben@niles ~ $ ./xcvradm -i axf2 power
FPGA Port Power Software override
   0    0 Off   -
   0    1 Off   -
   0    2 Off   -
   0    3 Off   -
   0    4 Off   -
   0    5 Off   -
   0    6 Off   -
   0    7 Off   -
   0    8 Off   -
   0    9 Off   -
   0   10 Off   -
   0   11 Off   -
   0   12 Off   -
   0   13 Off   -
   0   14 Off   -
   0   15 Off   -
```

I added a basic check that we have to go through low-power to get to high-power mode:

```bash
ben@niles ~ $ ./xcvradm -i axf2 -t 0 set-power high
Error: Failed to set power mode

Caused by:
    Invalid power state transition
ben@niles ~ $ ./xcvradm -i axf2 -t 0 set-power low
ben@niles ~ $ ./xcvradm -i axf2 -t 0 power
FPGA Port Power Software override
   0    0 Low   No
```

Then we can independently toggle the software control bit, without changing the actual power state:

```bash
ben@niles ~ $ ./xcvradm -i axf2 -t 0 power
FPGA Port Power Software override
   0    0 Low   No
ben@niles ~ $ echo '03' | ./xcvradm -iaxf2 -t0 write-lower --sff 93 --input-kind hex
ben@niles ~ $ ./xcvradm -i axf2 -t 0 power
FPGA Port Power Software override
   0    0 Low   Yes
```

Now the module is has set software override. Setting the power to high now deasserts the `LPMode` signal and also sets the right bit in the memory map:

```bash
ben@niles ~ $ ./xcvradm -i axf2 -t 0 set-power high
ben@niles ~ $ ./xcvradm -i axf2 -t 0 power
FPGA Port Power Software override
   0    0 High  Yes
ben@niles ~ $ ./xcvradm -i axf2 -t 0 status
FPGA Port Status
   0    0 PRESENT | ENABLED | INTERRUPT | POWER_GOOD
ben@niles ~ $ ./xcvradm -iaxf2 -t0 read-lower --sff 93 1
FPGA Port Data
   0    0 [0x01]
ben@niles ~ $
```

After turning _off_ software-override of power control (but not changing the power mode), we can also set the power back to low, which asserts `LPMode` and writes the correct bit:

```bash
ben@niles ~ $ ./xcvradm -i axf2 -t 0 power
FPGA Port Power Software override
   0    0 High  Yes
ben@niles ~ $ ./xcvradm -i axf2 -t 0 status
FPGA Port Status
   0    0 PRESENT | ENABLED | INTERRUPT | POWER_GOOD
ben@niles ~ $ ./xcvradm -iaxf2 -t0 read-lower --sff 93 1
FPGA Port Data
   0    0 [0x01]
ben@niles ~ $ echo '02' | ./xcvradm -iaxf2 -t0 write-lower --sff 93 --input-kind hex
ben@niles ~ $ ./xcvradm -i axf2 -t 0 power
FPGA Port Power Software override
   0    0 High  No
ben@niles ~ $ ./xcvradm -iaxf2 -t0 status
FPGA Port Status
   0    0 PRESENT | ENABLED | INTERRUPT | POWER_GOOD
ben@niles ~ $
```

bnaecker:
Resolves #41

controller/src/bin/xcvradm.rs
```diff
@@ -747,6 +770,23 @@ async fn address_transceivers(
 
 // Column width for printing data below.
 const WIDTH: usize = 4;
+const POWER_WIDTH: usize = 5;
+
+fn print_power_mode(modules: ModuleId, modes: Vec<(PowerMode, Option<bool>)>) {
+    println!("FPGA Port Power Software override");
```
mkeeter, line 776: Nit: it's not obvious that `Software override` is one column, versus two.  Maybe two spaces between columns?

controller/src/lib.rs
```diff
@@ -102,10 +103,13 @@ pub enum Error {
...
     )]
     InvalidInterfaceForModule(ManagementInterface),
+
+    #[error("Invalid power state transition")]
+    InvalidPowerStateTransition,
 }
 
 /// An allowed power mode for the module.
-#[derive(Clone, Debug, PartialEq, clap::ValueEnum)]
+#[derive(Clone, Copy, Debug, PartialEq, clap::ValueEnum)]
```
mkeeter, line 112: Are the comments below still correct?

e.g.
```rust
/// Note: This requires that we never set the `Power_override` bit (SFF-8636
/// rev 2.10a, section 6.2.6, byte 93 bit 2), as that defeats the purpose of
/// hardware control.
```

controller/src/lib.rs
```diff
@@ -508,68 +512,173 @@ impl Controller {
...
+            .copied()
+            .enumerate()
+            .filter_map(|(ix, st)| {
+                let index = u8::try_from(ix).expect("Impossible index");
+                if powered_modules.contains(index) {
+                    Some(st)
+                } else {
+                    None
+                }
+            })
```
mkeeter, line 596: Personal opinion: I'm not sold on `filter_map` being simpler than splitting up the filter and map, since you have to build the `Option` yourself...

(this also applies below)

```suggestion
            .filter(|(ix, _)| {
                let index = u8::try_from(ix).expect("Impossible index");
                powered_modules.contains(index)
            })
            .map(|(_, st)| st)
```

controller/src/lib.rs
```diff
@@ -632,6 +747,215 @@ impl Controller {
...
+            PowerMode::Off => {
+                self.assert_lpmode(modules).await?;
+                self.assert_reset(modules).await?;
+                self.disable_power(modules).await
+            }
+            other => {
+                // Validate the power state transition.
+                //
+                // For now, we enforce that modules may not go directly to high
+                // power, they have to go through low-lower first.
```
mkeeter, line 768:
```suggestion
                // power, they have to go through low-power first.
```

controller/src/lib.rs
```diff
@@ -632,6 +747,215 @@ impl Controller {
...
+                //
+                // We can always set modules to low power, though, since that's
+                // valid from both off and high-power, and a no-op if it's
+                // already set.
+                let current_power_state = self.power_mode(modules).await?;
+                if matches!(mode, PowerMode::High) {
+                    if current_power_state
```
mkeeter, line 775: Nit: this could be folded into the outer conditional

controller/src/lib.rs
```diff
@@ -632,6 +747,215 @@ impl Controller {
...
+        // How we proceed largely depends on two things: whether we're turning
+        // the power OFF entirely, and whether a module has set software
+        // override of the `LPMode` pin.
+        match mode {
+            PowerMode::Off => {
+                self.assert_lpmode(modules).await?;
+                self.assert_reset(modules).await?;
+                self.disable_power(modules).await
+            }
+            other => {
```
mkeeter, line 764: I'd be tempted to use the weird `@` syntax, to explicitly spell out the power modes captured by this branch:
```suggestion
            other @ PowerMode::Low | PowerMode::High => {
```

controller/src/lib.rs
```diff
@@ -632,6 +747,215 @@ impl Controller {
...
+                // module, only which hardware signals it responds to.
+                let need_power_enabled = Self::filter_modules_with(modules, &status, |st| {
+                    !st.contains(Status::POWER_GOOD | Status::ENABLED)
+                })?;
+                let need_reset_deasserted =
+                    Self::filter_modules_with(modules, &status, |st| st.contains(Status::RESET))?;
+                if need_power_enabled.selected_transceiver_count() > 0 {
+                    self.enable_power(need_power_enabled).await?;
+                }
+                if need_reset_deasserted.selected_transceiver_count() > 0 {
```
mkeeter, line 803: Is it possible for some transceivers to already be **not in reset**, but have power disabled?

I believe that would be a violation of our intended sequencing, but should we detect / warn about it / handle it specially?

decode/src/lib.rs
```diff
@@ -669,13 +669,84 @@ impl ParseFromModule for MemoryModel {
...
+pub struct PowerControl {
+    /// If true, the module is configured for software control of low power
+    /// mode. In this case, the `LPMode` pin has no effect.
+    pub software_override: bool,
+    /// If true, the module is set to low power mode by software. If `false`,
+    /// then the module is allowed to enter high power mode.
+    ///
+    /// This is _only meaningful_ if `software_override` is `true`.
+    pub low_power: bool,
+}
```
mkeeter, line 683: If `low_power` is only meaningful when `software_override` is true, that suggests this could be

```suggestion
pub enum PowerControl {
    /// The module is configured to use the pin to select low power mode
    UseLpModePin,

    /// The module is configured for software control of low power mode
    OverrideLpModePin {
        /// If true, the module is set to low power mode by software. If
        /// `false`, then the module is allowed to enter high power mode.
        low_power: bool,
    },
}
```

Dunno if it's worth making the resulting changes, but this seems like a more faithful model of what you're trying to encode.

controller/src/lib.rs
bnaecker: Nope, they're not! Good catch.

controller/src/lib.rs
bnaecker: Never had a reason to use that, but I like it here thanks!

decode/src/lib.rs
bnaecker: That does seem better, thanks!

controller/src/lib.rs
```diff
@@ -508,68 +512,173 @@ impl Controller {
...
+        // control is enabled.
+        let readable_modules = ModuleId {
+            fpga_id: modules.fpga_id,
+            ports: powered_modules.ports.remove(in_reset.ports),
+        };
+        let readable_status: Vec<_> = status
+            .iter()
+            .copied()
+            .enumerate()
+            .filter_map(|(ix, st)| {
```
Aaron-Hartwig, line 623: I think Matt's `filter_map` feedback above would apply here as well.

controller/src/lib.rs
```diff
@@ -632,6 +747,215 @@ impl Controller {
...
+    ///
+    /// This method may be used regardless of whether a module uses hardware
+    /// control or software override for controlling the power.
+    pub async fn set_power_mode(&self, modules: ModuleId, mode: PowerMode) -> Result<(), Error> {
+        // How we proceed largely depends on two things: whether we're turning
+        // the power OFF entirely, and whether a module has set software
+        // override of the `LPMode` pin.
+        match mode {
+            PowerMode::Off => {
+                self.assert_lpmode(modules).await?;
```
Aaron-Hartwig, line 760: While the module is powered off we actually want `LpMode` to be deasserted (low). The out of date RFD 244 has [the context](https://rfd.shared.oxide.computer/rfd/0244#_fpga_module_sequencing) which still holds:
>While we want to assert LPMode as soon as possible, we cannot do so until the module has had power enabled, otherwise the modules bleed out onto their main voltage rail. This is documented in [hardware-qsfp-x32#47](https://github.com/oxidecomputer/hardware-qsfp-x32/issues/47)

controller/src/lib.rs
Aaron-Hartwig: I think this is something we should check for since we there are timing requirements around reset, particularly the first reset after power on ("device initialization").

bnaecker: I would not be surprised if interactions with the SDE left a transceiver in that sort of state, though it'd be a bit weird. That would mean the SDE put a transceiver in reset, and then some of our software powered-off the module, using the direct interface like `xcvradm disable-power`.

I think it should be fairly easy to at least warn about this case, with a few bit operations on the `PortMask`s. I'll try it out, and either add it now if it's easy or make an issue. Thanks for bringing it up!

controller/src/lib.rs
bnaecker: Ok, thanks for the context. I'll change that and add a link.

So I understand, does this mean that it's possible for a module to turn on and go directly to high-power mode? The code in this PR always turns on the power then sets `LPMode`, but those could be separated by a while since this is all over the network. What happens in that case?

controller/src/lib.rs
bnaecker: @mkeeter @Aaron-Hartwig Fixed this up in 4cfceaf. I've made this so that, if we are enabling power on any modules which _do not_ already have `ResetL` asserted, we:

- Emit a warning
- Assert `ResetL`
- Wait for SFF-8679 `t_reset_init` (10us)
- For all modules being powered (even those which previously did not have `ResetL` asserted, i.e., the odd ones), then assert `ResetL`
- Wait for SFF-8679 `t_reset` (2s)

bnaecker: That behavior now looks like this:

```bash
ben@niles ~ $ ./xcvradm -i axf2 -t0 status
FPGA Port Status
   0    0 PRESENT | INTERRUPT
ben@niles ~ $ ./xcvradm -i axf2 -t0 set-power low
Jan 31 20:59:46.840 WARN Found modules with power disabled, but reset deasserted. It will be asserted before enabling power., suspicious_modules: ModuleId { fpga_id: 0, ports: PortMask(1) }, need_reset_deasserted: ModuleId { fpga_id: 0, ports: PortMask(0) }, need_power_enabled: ModuleId { fpga_id: 0, ports: PortMask(1) }
ben@niles ~ $ ./xcvradm -i axf2 -t0 status
FPGA Port Status
   0    0 PRESENT | ENABLED | LOW_POWER_MODE | INTERRUPT | POWER_GOOD
ben@niles ~ $ ./xcvradm -i axf2 -t0 power
FPGA  Port  Power  Software-override
   0     0  Low    No
ben@niles ~ $
```

controller/src/lib.rs
```diff
@@ -757,27 +754,33 @@ impl Controller {
...
+                // https://github.com/oxidecomputer/hardware-qsfp-x32/issues/47
+                // for the hardware issue and
+                // https://rfd.shared.oxide.computer/rfd/0244#_fpga_module_sequencing
+                // for a general discussion.
+                self.deassert_lpmode(modules).await?;
                 self.assert_reset(modules).await?;
                 self.disable_power(modules).await
             }
-            other => {
+            new_mode @ PowerMode::Low | new_mode @ PowerMode::High => {
```
mkeeter, line 768: Nit-picking my own previous suggestion, I just realized that we already have this variable captured in `mode`, so we don't actually need `new_mode` at all here!
```suggestion
           PowerMode::Low | PowerMode::High => {
```

bnaecker: Ha true enough! Thanks.

mkeeter (approved):

controller/src/lib.rs
```diff
@@ -632,6 +744,253 @@ impl Controller {
...
+                        .remove(&need_reset_deasserted.ports),
+                };
+                if need_power_but_not_reset.selected_transceiver_count() > 0 {
+                    warn!(
+                        self.log,
+                        "Found modules with power disabled, but reset deasserted. \
+                        It will be asserted before enabling power.";
+                        "need_power_enabled" => ?need_power_enabled,
+                        "need_reset_deasserted" => ?need_reset_deasserted,
+                        "suspicious_modules" => ?need_power_but_not_reset,
```
Aaron-Hartwig, line 820: Nit: This ends up printing as `.,` in `enabling power., $` when any of those cases evaluate to true.

controller/src/lib.rs
```diff
@@ -632,6 +744,253 @@ impl Controller {
...
+                    self.deassert_reset(need_power_enabled).await?;
+
+                    // The SFF-8769 specifies that modules may take up to 2
+                    // seconds after asserting ResetL before they are ready for
+                    // reads. This is `t_reset`.
+                    tokio::time::sleep(Duration::from_secs(2)).await;
+                }
+
+                // Actually write the power mode, the pin and to the memory map.
+                self.set_lp_mode(modules, &current_power_state, mode).await
```
Aaron-Hartwig, line 848: To be extra careful we could put this between the `enable_power` and `deassert_reset` above. This ensures that we are asserting `LpMode` _after_ applying power (addressing the voltage problem mentioned above) but _before_ we take the device out of reset. This is definitely belt and suspenders, but figured I'd mention it.

Aaron-Hartwig (approved):
LGTM. Left a few comments I don't think are critical fix.

controller/src/lib.rs
bnaecker: Yeah, `slog` auto-inserts `,` between every "element" of the log message. I can remove it.

controller/src/lib.rs
bnaecker: Ok, I can do that. It's a bit of reworking, but I already have the methods for it. It looks a bit funky, because we set the hardware / software `LPMode` before and after `deassert_reset`, but I think that only matters when there are modules that we're bringing from off -> low-power. For those going from high -> low or low -> high, those calls to `enable_power` and `assert_reset` don't apply, since those modules aren't in the mask `need_power_enabled`.

Thanks, will update now.

bnaecker: Done in 8c46545, let me know what you think @Aaron-Hartwig.

Aaron-Hartwig: Looking at that diff, I think I better understand the "weirdness" around how we have to manage both the hardware and software state involved with low vs high power mode.  https://github.com/oxidecomputer/transceiver-control/commit/8c46545ba16416408f6d3fe498e311d38a962a20 gives what I asked for and definitely is not as "clean" as the previous implementation, but I do think it is a more accurate reflection of what is actually going on.

bnaecker: Great, thanks for helping make that accurate! I agree that it's a bit confusing, since we're interspersing the hardware / software mode control between twiddling the power and reset hardware signals. I tried to lay the reasoning out clearly in the block comment, and hopefully neither of us is confused the next time we come back here!

PR #58, opened 2023-02-08
Add message for collecting MAC addresses from SP

bnaecker:
- Add the `MacAddrs` message to the protocol, and `MacAddrs` type for representing the MAC range for various kinds of systems.
- Adds method to `Controller` and `xcvradm` subcommand for exercising.

### DIFF
--- a/messages/src/lib.rs
+++ b/messages/src/lib.rs
@@ -152,12 +152,15 @@ pub enum Error {
         any(test, feature = "std"),
         error(
             "The MAC address range is invalid, \
-            reason: {reason}, range: {range:?}"
+            reason: {reason}, base: {base_mac:?}, \
+            count: {count}, stride: {stride}"
         )
     )]
     BadMacAddrRange {
         reason: BadMacAddrReason,
-        range: MacAddrs,
+        base_mac: [u8; 6],
+        count: u16,
+        stride: u8,
     },
 }
 
@@ -505,15 +508,12 @@ impl MacAddrs {
     /// is not a multicast address, that all addresses in the range share the
     /// same OUI, and that count/stride are nonzero.
     pub const fn new(base_mac: [u8; 6], count: u16, stride: u8) -> Result<Self, Error> {
-        let range = Self {
-            base_mac,
-            count,
-            stride,
-        };
         if count == 0 || stride == 0 {
             return Err(Error::BadMacAddrRange {
                 reason: BadMacAddrReason::ZeroStrideOrCount,
-                range,
+                base_mac,
+                count,
+                stride,
             });
         }
 
@@ -521,20 +521,27 @@ impl MacAddrs {
         if (base_mac[0] & 0b1) != 0 {
             return Err(Error::BadMacAddrRange {
                 reason: BadMacAddrReason::MulticastBaseAddr,
-                range,
+                base_mac,
+                count,
+                stride,
             });
         }
 
         // Check that the last address, at `n * stride` does not change the OUI.
         let base_n = u32::from_be_bytes([base_mac[2], base_mac[3], base_mac[4], base_mac[5]]);
         let n = count as u32;
-        let stride = stride as u32;
-        let offset = n * stride; // Cannot overflow based on types.
+        let offset = n * stride as u32; // Cannot overflow based on types.
         match base_n.checked_add(offset) {
-            Some(last) if last.to_be_bytes()[0] == base_mac[2] => Ok(range),
+            Some(last) if last.to_be_bytes()[0] == base_mac[2] => Ok(Self {
+                base_mac,
+                count,
+                stride,
+            }),
             _ => Err(Error::BadMacAddrRange {
                 reason: BadMacAddrReason::SpansMultipleOuis,
-                range,
+                base_mac,
+                count,
+                stride,
             }),
         }
     }
--- a/messages/src/lib.rs
+++ b/messages/src/lib.rs
@@ -450,19 +450,58 @@ impl ModuleId {
 /// The MAC address allocation for the target system.
 #[derive(Clone, Copy, Debug, Deserialize, Eq, PartialEq, Serialize, SerializedSize)]
 pub struct MacAddrs {
-    /// The first valid MAC address for the system.
-    pub base_mac: [u8; 6],
-    /// The number of available addresses following `base_mac`.
-    pub count: u16,
-    /// The stride pattern used to compute the next available MAC address.
-    ///
-    /// This is used in situations in which subsystems may make assumptions
-    /// about spacing between addresses. At the time of writing, this applies to
-    /// the Chelsio T6 NICs.
-    pub stride: u8,
+    // The first valid MAC address for the system.
+    base_mac: [u8; 6],
+    // The number of available addresses following `base_mac`.
+    count: u16,
+    // The stride pattern used to compute the next available MAC address.
+    //
+    // This is used in situations in which subsystems may make assumptions
+    // about spacing between addresses. At the time of writing, this applies to
+    // the Chelsio T6 NICs.
+    stride: u8,
 }
 
 impl MacAddrs {
+    /// Create a new MAC address range.
+    ///
+    /// This will do some basic sanity checks on the provided data to make sure
+    /// that the full range is valid. Specifically, we check that the base MAC
+    /// is not a multicast address, that all addresses in the range share the
+    /// same OUI, and that count/stride are nonzero.
+    pub const fn new(base_mac: [u8; 6], count: u16, stride: u8) -> Option<Self> {
+        if Self::is_valid(base_mac, count, stride) {
+            Some(Self {
+                base_mac,
+                count,
+                stride,
+            })
+        } else {
+            None
+        }
+    }
+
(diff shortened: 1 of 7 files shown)


bnaecker:
This is a WIP, and should not be merged until we implement the SP side as well, since it bumps the protocol version.

mkeeter:
@bnaecker Can you rebase this and bump the version again (since #60 already bumped it to V5)?  Once that's done, I'll be ready to do the Hubris implementation.

messages/src/message.rs
```diff
@@ -130,6 +133,9 @@ pub enum HostRequest {
     /// transceivers' memory maps such as the temperature or power draw, for the
     /// purposes of health and safety monitoring.
     ManagementInterface(ManagementInterface),
+
+    /// Ask the SP to return the available MAC addresses for host system use.
+    MacAddrs(MacAddrKind),
```
mkeeter, line 138: I'm not sure including the `MacAddrKind` here makes sense: the SP technically knows whether it's a Gimlet / Sidecar / PSC, but it only has one block of MAC addresses to give in every case.

bnaecker: Yeah, I wasn't sure whether we needed this. Just to make sure I understand, this particular request, which is from the Scrimlet host to the Sidecar SP, always returns the Sidecar block of addresses, i.e, (base, count = 256, stride = 1). Is that accurate? I guess I wondered if any other software would use this same mechanism, but since it's the transceivers task handling the request, which only runs on the Sidecar SP, then seems like the answer is "no".

If that's all accurate, I'll remove this when I bring this up to date with `main`.

mkeeter: Good point that this is only answered by the `transceivers` task!

Yes, asking the Sidecar SP will only return the Sidecar's addresses – or rather, it will return whatever is in the VPD EEPROM (after the SP itself claims the first two addresses).  We're trusting that the VPD is configured correctly for the board that it's installed on.

bnaecker: @mkeeter I rebased on `main` and removed the kind bits. Let me know how the Hubris side goes, if we need to update this at all!

messages/src/lib.rs
```diff
@@ -426,16 +431,87 @@ impl ModuleId {
...
+
+    /// Return the `n`th address in the range of MACs provided by `self`.
+    ///
+    /// If `n` is outside of the range of `self`, `None` is returned.
+    pub fn nth(&self, n: u16) -> Option<[u8; 6]> {
+        if n >= self.count {
+            return None;
+        }
+        let base_n = u16::from_be_bytes([self.base_mac[4], self.base_mac[5]]);
+        let shift = (base_n + n * u16::from(self.stride)).to_be_bytes();
```
mkeeter, line 482: How about a `checked_add` here to catch avoid potential overflow?

mkeeter (approved):

messages/src/lib.rs
bnaecker: Nice, will do

bnaecker: I guess a `checked_mul` would be helpful too :)

bnaecker: Good thing I looked again. This implementation is not correct. I believe it basically assumes that the base address has zero for the last two octets.

bnaecker: I guess this does raise the question whether we should check that the resulting MAC always has the Oxide OUI. It seems like should _not_ allow MACs without that, but I'm not sure what to do in that case.

It seems odd, but we could get a base MAC like `a8:40:25:ff:ff:00`. A count of say 257 or more would then result in MACs like `a8:40:26:....`, which seems like something we should avoid.

rmustacc: While checking that you don't exceed the OUI is useful, I don't think it's appropriate to hard code the list of Oxide OUIs. We had similar discussions around this with @citrus-it and others when we're looking at this in the context of Hubris and the host OS. The valid set of OUIs we have will change over time and it seems like we shouldn't rely on a hardware update for that. The block overlap is an important part and something we need to track in manufacturing assignment (cc @jclulow).

bnaecker: Sounds good. Current plan is to keep things within the _existing_ OUI in `base_mac` when computing the `nth` address or in the iterator, but not checking anything about the specific OUI.

jclulow: FWIW in manufacturing we are always assigning 256 addresses for sidecars; each base address is a round number of the form `OO:UU:II:aa:bb:00`.

bnaecker: Talked more with Robert here. It seems like a better plan is to fail pretty hard, pretty fast. We'll do some checks in the constructor that the full range is realizable, and if we don't pass some basic sanity checks then we fail right away.

messages/src/lib.rs
```diff
@@ -426,16 +431,144 @@ impl ModuleId {
...
+}
+
+impl MacAddrs {
+    /// Create a new MAC address range.
+    ///
+    /// This will do some basic sanity checks on the provided data to make sure
+    /// that the full range is valid. Specifically, we check that the base MAC
+    /// is not a multicast address, that all addresses in the range share the
+    /// same OUI, and that count/stride are nonzero.
+    pub const fn new(base_mac: [u8; 6], count: u16, stride: u8) -> Option<Self> {
```
mkeeter, line 472: Since this now returns an `Option`, can you add a `BadMacAddrBlock` to the [`enum Error`](https://github.com/oxidecomputer/transceiver-control/blob/main/messages/src/lib.rs#L43) above?  This will allow Hubris to report a meaningful error if it's got bad data (instead of unwrapping it).

bnaecker:
Yeah, it occurred to me after the fact that this would mean hubris has to deal with the fallibility of the constructor. I’d assumed we’d want to make the constructor infallible and do the validation explicitly on the host side, but your suggestion of a new error variant seems better. Thanks, I’ll do that this morning!

messages/src/lib.rs
Aaron-Hartwig: I would vote for including the bad data as "what was the bad address" is the first question one would ask I'd think. Closely followed by "how did we generate it". Being able to jump right into the next step of debugging seems valuable.

bnaecker: Wait I think I misunderstood your suggestion, @Aaron-Hartwig. You mean include the exact base / stride / count in the error, rather than a description of the error?

messages/src/lib.rs
```diff
@@ -469,26 +498,25 @@ impl MacAddrs {
     /// that the full range is valid. Specifically, we check that the base MAC
     /// is not a multicast address, that all addresses in the range share the
     /// same OUI, and that count/stride are nonzero.
-    pub const fn new(base_mac: [u8; 6], count: u16, stride: u8) -> Option<Self> {
-        if Self::is_valid(base_mac, count, stride) {
-            Some(Self {
+    pub const fn new(base_mac: [u8; 6], count: u16, stride: u8) -> Result<Self, Error> {
```
mkeeter, line 501: API musings: since this function will only ever return `Error::BadMacAddrBlock(...)`, maybe have it return the inner `BadMacAddrReason` directly (and let the caller wrap it if desired)?

messages/src/lib.rs
```diff
@@ -469,26 +498,25 @@ impl MacAddrs {
...
+        match Self::is_valid(base_mac, count, stride) {
```
mkeeter, line 502: This could be
```rust
Self::is_valid(base_mac, count, stride).map(|_| Self { base_max, count, stride })
```
or even
```rust
Self::is_valid(base_mac, count, stride)?;
Ok(Self { base_max, count, stride }))
```

I don't have strong opinions on which of the three is preferable.

mkeeter: (or even folding the validity checks into the constructor, since `is_valid` is not public!)

messages/src/lib.rs
Aaron-Hartwig: I think it should be the bad data (base + stride + count) _with_ a description (?) rather than just a description on its own.

messages/src/lib.rs
bnaecker: Yeah, I played with that a bunch. Those forms are not const, though, so the constructor can't be. That's not the end of the of the world, of course. I don't really feel strongly either way.

messages/src/lib.rs
bnaecker: That maybe seems fine? Talking through that though, the `transceivers` task would need to wrap it immediately in order to send it to the host, since it has to be stuffed into an `SpResponse::Error` variant. On the host-side we're just going to deserialize it directly, so I don't think this code even runs, right? So I guess it's just up to you whether you'd like include that (minimal) logic in the Hubris task.

messages/src/lib.rs
bnaecker: Folding the checks in makes this problem go away :) Thanks for the suggestion!

messages/src/lib.rs
bnaecker: I've reworked this so the `Error::BadMacAddrRange` variant now contains the reason _and_ the offending data. So the simplification here would require adding a new type to store both, and then using that in the `Error` variant.

mkeeter: Okay, that seems good to me.

mkeeter (approved):

Issue #63, opened 2023-02-24
Add derives for using some data in Dropshot APIs


bnaecker:
There are a few types that we'd like to use in Dropshot APIs. This will be at least Dendrite, but some of these may also be exported all the way to the internal or even external Nexus API as well. We should add a feature (with associated optional dependencies) for deriving `serde::{Deserialize, Serialize}` and `schemars::JsonSchema` for the following types:

- [x] `PowerMode`: This is the primary way folks will view or control the power of a transceiver
- [x] `VendorInfo`: The primary identifying information for a transceiver. It'll be most notably used to derive topology information and unique identifiers for each transceiver. See [RFD 360](https://rfd.shared.oxide.computer/rfd/0360#_qsfp_transceivers).
- [x] Health and monitoring data: These don;t exist yet in `transceiver-decode`, but will need to soon, and will need to be visible via Dendrite's API. This includes: (1) current / voltage; (2) per-lane optical power; (3) cage temperature; (4) possibly alarm or fault state

It's not clear what we should do with `Status`. I'm inclined to think this should be changed from a bitmap to a struct, showing the information per-transceiver e.g., as boolean fields.

Aaron-Hartwig:
>It's not clear what we should do with Status. I'm inclined to think this should be changed from a bitmap to a struct, showing the information per-transceiver e.g., as boolean fields.

That seems logical to me. What would keeping it as a bitmap get us?

bnaecker:
I dunno. I was wondering if we would serialize it as a string, which I think `bitflags` does like `FOO | BAR | BAZ`. That's not super helpful, so a struct seems the way to go. I'll play around with it.

bnaecker:
Completed by #68

PR #64, opened 2023-02-24
Add derivations of API-related traits

bnaecker:
- Derives `serde::Serialize`, `serde::Deserialize` and `schemars::JsonSchema` for `transceiver_control::PowerMode`.
- Derives `serde::Serialize`, `serde::Deserialize` and `schemars::JsonSchema` for `transceiver_decode::Vendor` and all required sub-types. This means serializing the OUI as a formatted string for now.
- Removes any parsing of the Date Code at all. It's either kept as an ASCII string (which the spec mandates it's supposed to be) or None, and printed that way. This avoids all the annoyance around parsing it into a type, which we really only care about at this point for printing anyway. This helps serialization, since it's just a string or nothing.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,15 +3,6 @@
 version = 3
 
 [[package]]
-name = "android_system_properties"
-version = "0.1.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


bnaecker:
Partially resolves #63

Aaron-Hartwig (approved):

mkeeter (approved):

PR #66, opened 2023-02-26
Add committed minimum protocol version

bnaecker:
- Adds a minimum version of the network protocol. This is _committed_, in that we can never break backwards compatibility. There is also a CURRENT version, which can be different. This only serves as informational, to detect version mismatches and possibly warn about them.
- Adds tests that we can correctly [de]serialize _all_ current data used in the protocol. This should catch any changes that break the existing wire-encoding.
- Adds a sanity check for the new behavior in the IO loop for handling responses from the SP that do not match our version. We currently only _fail_ responses below our minimum committed version. We should never get responses to these, since all existing SP code will drop messages that don't match its own version, but this is a helpful sanity check. We also add basic tests that we still correctly receive, decode, and handle responses from the SP where the version is below our own, but above the minimum, and also above our own. Assuming that the SP is also updated to respect this protocol, we should not have any further protocol mismatch errors _that we cannot detect on the host_. Specifically, the SP should be updated to try to handle host requests, even when those don't match its own version. If it can successfully decode the message, it should send a response (success or error). If it _cannot_ decode the message, it should send a protocol error or other error indicating that. Those error messages _are_ among the currently-supported set, and so should always be receivable by the host.

### DIFF
--- a/controller/src/lib.rs
+++ b/controller/src/lib.rs
@@ -1607,30 +1607,65 @@ impl IoLoop {
                     );
                     probes::message__received!(|| (peer.ip(), &message));
 
-                    // Sanity check the protocol version.
-                    if message.header.version != message::version::CURRENT {
-                        // If the version does not match, we're choosing to drop
-                        // the packet rather than reply with a version mismatch
-                        // error. Without a matching version, we can't really
-                        // trust the message kind we have deserialized, so won't
-                        // be able to reliably send protocol errors.
+                    // Check the version in the response.
+                    //
+                    // Beginning at version `MIN` (V5), we committed to
+                    // backwards compatibility. In particular, any existing
+                    // messages must still be decodable by software running a
+                    // later version of the protocol.
+                    //
+                    // In this case we're processing a response to our own
+                    // request. If we _get_ a response at all, it means that the
+                    // peer is running at least `MIN`. There are actually no
+                    // other checks that need to be performed, which would
+                    // prevent us from decoding the message.
+                    //
+                    // Suppose the peer is running between `MIN` and `CURRENT`.
+                    // Then we can clearly process their response, because the
+                    // version is one that we've committed to compatibility
+                    // with. This could be a version-mismatch error message,
+                    // because the _peer_ may not be able to handle the message.
+                    //
+                    // If the version is _newer_ than `CURRENT`, we can still
+                    // process it. That's because all of our messages can be
+                    // decoded and processed by the peer, who has also committed
+                    // to this compatibility.
+                    if message.header.version < message::version::MIN {
                         debug!(
                             self.log,
-                            "deserialized message with incorrect version";
-                            "expected" => message::version::CURRENT,
-                            "actual" => message.header.version,
+                            "deserialized message with incompatible version";
+                            "current" => message::version::CURRENT,
+                            "min" => message::version::MIN,
+                            "message_version" => message.header.version,
                             "peer" => peer,
                         );
                         probes::bad__message!(|| {
                             (
                                 peer.ip(),
                                 format!(
-                                    "incorrect version: expected {}, actual {}",
+                                    "incompatible version: \
+                                    current {}, min {}, message_version {}",
                                     message::version::CURRENT,
+                                    message::version::MIN,
                                     message.header.version,
                                 ),
                             )
                         });
+
+                        // In this case, we're never going to be able to process
+                        // a response from the peer. If this is a response to
+                        // our outstanding message; send an error message on its
+                        // response channel; and throw it away. If this is a
+                        // message from the peer, just do nothing, since they
+                        // won't be able to handle our response anyway.
+                        if let Some(request) = self.outstanding_request.take() {
+                            request.response_tx.send(
+                                Err(Error::Protocol(MessageError::VersionMismatch {
+                                    expected: message::version::CURRENT,
+                                    actual: message.header.version
+                                }))
+                            ).unwrap();
+                        }
                         continue;
                     }
 
@@ -1831,12 +1866,32 @@ impl IoLoop {
 #[cfg(test)]
 mod tests {
     use super::is_link_local;
+    use super::message;
+    use super::mpsc;
+    use super::oneshot;
     use super::sff8636;
     use super::verify_ids_for_page;
     use super::ConfigBuilder;
+    use super::Duration;
+    use super::Error;
+    use super::Header;
+    use super::HostRequest;
+    use super::HostRpcRequest;
     use super::Identifier;
+    use super::IoLoop;
+    use super::Logger;
+    use super::Message;
+    use super::MessageBody;
+    use super::MessageError;
+    use super::ModuleId;
+    use super::OutstandingHostRequest;
     use super::Page;
     use super::PowerMode;
+    use super::SocketAddr;
+    use super::SocketAddrV6;
+    use super::SpResponse;
+    use super::SpRpcResponse;
+    use super::UdpSocket;
     use std::net::Ipv6Addr;
 
     #[test]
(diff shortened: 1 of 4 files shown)


bnaecker:
This is a bit of a strawman. But we are intending to add new messages to the protocol soon to support LED control, and I'd like to get some eyes on this before we do that. I'm adding a _committed_ minimum protocol version. If we agree to this, it will mean that any changes must be additive, and not change existing message data structures in a way that would break serialization. This would mean that peers should _try_ to decode and respond to messages, even if they are from a different protocol version, as long as they meet the minimum. Because the `Error::VersionMismatch` is part of this, we should always be able to detect when the peer is of such a different version that they cannot handle our message.

I've added a bunch of comments and tests. The first set of tests checks the existing `hubpack` encodings of all protocol data types. This will catch any changes that break existing encoding, and we should add tests for new messages as we add them (but not change the existing ones!). The second does a few sanity checks of the code that actually checks the protocol version from the SP. That is a bit dumb, in that we're just catching (1) a version below the minimum, which we should never get because all SP code currently drops any mismatching message, and (2) that we _can_ successfully decode messages from a older or newer protocol version from our own. I think it would be good to test this more rigorously when we add a _new_ message type (say the LED stuff). We can simulate the SP being too old to handle that new message, and make sure that the host does correctly deserialize an `Error::VersionMismatch`. We should of course update the Hubris code to act this way too!

cc @jclulow We discussed this a bunch in chat. I appreciate those comments immensely. This PR basically implements what we arrived at, but I'd be happy to take your suggestions based on the actual code.

bnaecker:
So I've just done some testing on this. I implemented the Hubris side [here](https://github.com/oxidecomputer/hubris/tree/xcvr-protocol-version-range), built that and flashed it to `niles/sidecar-b`. The SP was originally running this code:

```bash
ben@niles ~ $ pfexec humility --target sidecar-b manifest
     version => hubris build archive v7
     git rev => 0fff257b362c82c6b34a4cfe3812dd9224360b3a-dirty
    image id => [eb, 17, b0, 58, b5, cc, ef, 59]
       board => sidecar-b
        name => sidecar-b-lab
...
```

I ran a test using both `xcvradm` from `main` (fc6522d62a8de67b3f627113fe4b11fa03c99d56) and from the [`led-control` branch](https://github.com/oxidecomputer/transceiver-control/tree/led-control) (currently at 343666db87c38ebb5fd69b60b965aa9cf488ca18). I used that branch rather than this PR specifically because that bumped the protocol version again, since we added messages about the LEDs. This was to exercise the part of the Hubris code that checks against the _minimum committed_ version and a version that's _ahead_ of the Hubris version. Here's what I see.

When running `xcvradm` from `main` and the Hubris bits above, everything works fine.

```bash
ben@niles ~ $ ./xcvradm.orig -i axf2 -t 0 -n 1 status
FPGA Port Status
   0    0 PRESENT | ENABLED | LOW_POWER_MODE | INTERRUPT | POWER_GOOD
```

If I keep the Hubris code at that version, but bump the `xcvradm` to the `led-control` branch, which is ahead of the SP's `transceivers` task, then I get no response at all:

```bash
ben@niles ~ $ ./xcvradm -i axf2 -t 0 -n 1 status
Feb 26 23:01:00.931 ERRO failed to send message within 1 retries, task: io
Error: Failed to retrieve module status

Caused by:
    Maximum number of retries (1) reached without a response
ben@niles ~ $ pfexec humility --target sidecar-b ringbuf transceivers
humility: attached to 0483:3754:002600184D4B500E20373831 via ST-Link V3
humility: ring buffer drv_transceivers_server::__RINGBUF in transceivers:
 NDX LINE      GEN    COUNT PAYLOAD
   0  574        1        1 FrontIOReady(true)
   1  612        1        1 LEDInit
   2  140        1        1 LEDInitComplete
   3  552        1        1 ModulePresenceUpdate(0x5155557)
   4  258        1        1 GotInterface(0x0, Sff8636)
   5  258        1        1 GotInterface(0x2, Sff8636)
humility: ring buffer drv_transceivers_server::udp::__RINGBUF in transceivers:
 NDX LINE      GEN    COUNT PAYLOAD
   0  264        1        2 Status(ModuleId { fpga_id: 0x0, ports: PortMask(0xffff) })
   1  264        1        1 Status(ModuleId { fpga_id: 0x0, ports: PortMask(0x1) })
   2  202        1        1 WrongVersion(0x7)
ben@niles ~ $
```

We can see that this is because the Hubris code is logging a `WrongVersion` ringbuf entry, and then dropping the message. That's expected.

I then bumped the Hubris code by flashing the new bits. The SP is now running:

```bash
ben@niles ~ $ pfexec humility --archive build-sidecar-b-lab.zip --target sidecar-b manifest
humility: WARNING: archive on command-line overriding archive in environment file
     version => hubris build archive v7
     git rev => 78ed4a76d30d1f7d5d65a526311184aea6b7b321-dirty
    image id => [7d, ac, d5, 93, f4, 30, 9f, fc]
       board => sidecar-b
        name => sidecar-b-lab
      target => thumbv7em-none-eabihf
...
```

Now, both the `xcvradm` from `main` and `led-control` branches work fine.

```bash
ben@niles ~ $ ./xcvradm.orig -i axf2 -t 0 -n 1 status
FPGA Port Status
   0    0 PRESENT | ENABLED | LOW_POWER_MODE | INTERRUPT | POWER_GOOD
ben@niles ~ $ ./xcvradm -i axf2 -t 0 -n 1 status
FPGA Port Status
   0    0 PRESENT | ENABLED | LOW_POWER_MODE | INTERRUPT | POWER_GOOD
```

There's another interesting test. The Hubris code is now pretty lenient with protocol versions. The exact code is in the `xcvr-protocol-version-range` branch, but the behavior is this:

- Try to deserialize the message.
- If that fails, and the version is below the minimum committed, drop it and log a ringbuf entry.
- If that succeeds, and the version is below the minimum committed, we also drop it, on the theory that we really can't trust anything. This is the current behavior for any version that does not match our own.
- If deserialization fails and the version is _above_ min-committed, then this implies that the host-side came from a newer protocol version than our own. (It doesn't explicitly check this, which I should add.) Send back a `VersionMismatch` error.
- If deserialization succeeds, and the version is above min-committed, then we're assuming we can correctly handle the message. This relies on us not messing up backwards compatibility.

So, when I send a message from the _new_ `xcvradm` for a message that the SP should understand, things are fine. That's shown above. But when I send it a message that it _doesn't_ have, because the host-side is running a new protocol, we see this:

```bash
ben@niles ~ $ ./xcvradm -i axf2 -t 0 -n 1 leds
Error: Failed to get LED state

Caused by:
    0: Network protocol error
    1: Version mismatch: expected=6, actual=7
```

That is, the version-mismatch error here means that the host is newer and sending us a message we can't handle. It _might_ be worth adding an `Error::OperationUnsupported` variant, but that's really what the `VersionMismatch` means now anyway. So I'm not sure if that's important.

The summary is:

- Messages below min-committed are dropped.
- Messages that fail to deserialize, but are above the committed version must come from a _newer_ version than the peer, and generate an `Error::VersionMismatch` in response.
- Messages that deserialize correctly, and have a version above min-committed, are operated on normally.

Aaron-Hartwig (approved):
I think this and the Hubris PR look good. Thanks for working this out and documenting it so well, this is going to be very useful!

controller/src/lib.rs
```diff
@@ -1602,30 +1602,65 @@ impl IoLoop {
...
+                    // other checks that need to be performed, which would
+                    // prevent us from decoding the message.
+                    //
+                    // Suppose the peer is running between `MIN` and `CURRENT`.
+                    // Then we can clearly process their response, because the
+                    // version is one that we've committed to compatibility
+                    // with. This could be a version-mismatch error message,
+                    // because the _peer_ may not be able to handle the message.
+                    //
+                    // If the version is _newer_ than `CURRENT`, we can still
```
mkeeter, line 1624: This sentence is confusing to me – if the peer is running a version that's far ahead of `CURRENT`, why can we still process their messages?  Is it because at this point, we've already bailed out with deserialization errors if we couldn't parse it?

mkeeter (commented):
This generally looks good!

Small note: it's missing tests in `message.rs` for `Page` (all flavors) and `ManagementInterface` encoding.  The 0th value of these enums is checked implicitly (since the buffer is filled with 0s); however, if I maliciously added a new `ManagementInterface` between `Sff8636` and `Cmis`, it would break the encoding of the latter and not be detected by the test suite.

bnaecker:
Thanks for the eyes @mkeeter! I'll add tests for those types you mention, that was just an oversight on my part.

controller/src/lib.rs
bnaecker: I think that's right. And I also think this points out an issue. But first, let me expand on this part. Here, we've received a response to our request, which we can completely deserialize. The Sad Case is when that has a version before the compatibility commitment. We should never really get that, because any Hubris code running those versions will just drop a message whose version doesn't match its own, but I'm pretty paranoid at this point.

The point I was trying to make here is that it doesn't matter what version the peer is running, as long as its at least `MIN`. If they're way ahead of `CURRENT`, they still have to be able to process and respond to all the messages that _the host side_ can send. That's because the SP's current is a superset of ours. At least, it is if we correctly maintain backwards compatibility.

I hope that helps, but let me know how you think the comment can be improved. Some of this was written as I was solidifying the ideas, so I'm sure it can be tightened up.

---

Now to the problem this raises. If you look above, I'm handling deserialization failures in the same way as before: log and drop the message, but otherwise just continue. I'm not sure that's correct. In particular, I think we'll still _retry the message_. We need the same kind of error-injection I have in the code just below your comment, starting at L1656 in the PR diff. I.e., we need to `take()` the outstanding message, and send an error on the reply channel, so that the code making the request can bubble that up. Does that sound accurate?

bnaecker:
Tests added in c209dec

bnaecker:
I've resolved the conflicts locally, once there's approval of the latest changes, I'll force-push that and then squash+merge when CI is green.

controller/src/lib.rs
mkeeter: What about something like this?
```rust
                    // If the version is _newer_ than `CURRENT`, we can still
                    // process it, because we successfully deserialized it
                    // (above).  Our peer has committed to this same
                    // compatibility, so we know that successful deserialization
                    // means that we share a common understanding of the message
                    // semantics with our peer.
```

The retry issue was pre-existing before this PR, right?  If that's the case, then I agree with your assessment.

bnaecker: Yes, the retry issue is separate. I've updated the code to inject a failure up to the caller. I like your suggested comment, I'll work that in. Thanks!

PR #71, opened 2023-03-09
Support success and failure for all operations

bnaecker:
This commit adds support for returning information about both successes and failures for every operation. This is a fundamental redesign of the networking protocol, and thus breaks compatibility.

The core change is to add a module bitmask for both successes and failures in every response from the SP. The _cause_ for the failures are present in the trailing data of the UDP packet. Should the message contain both success and error data, the success data appears first, followed by the trailing data.

This has a huge blast radius. The network protocol has been rewritten. Most `Controller` methods are the same, but they return different types, which now include the successful modules and data, and failed modules and errors. Some highlights:

- The host side now addresses _all_ modules on both FPGAs, rather than per-FPGA as previously. That means the `ModuleId` type is a logical bitmask, where the indices correspond to the same port numbers one sees on the front of the Sidecar panel. The SP handles the mapping to the right FPGA internally.
- All responses from the SP include a `ModuleId` for success and failures.
- The error types have been restructured to be more modular and supportive of the error reporting from the SP.
- A number of result types have been implemented. These are all `ModuleResult`s or a type alias of it. This includes the successful modules; the set of data associated with those successes; and the failed modules and their failure causes.
- A number of convenience methods for filtering out `ModuleId`s and an associated array of data has been added. That includes the `FailedModules` type and all the `ModuleResult`s. This is very useful for handling errors during an operation which requires several requests to the SP. The modules that have failed can be pulled out and their failures recorded, at each stage. Further stages only operate on the modules that have succeeded at the prior stage.
- Adds much more rigorous testing. The core `IoLoop` type which handles the network now has a bunch of tests around parsing and validating messages both in and out. The new `ModuleId` type and result types all have tests around the filtering logic. The `Controller` itself also has a bunch of tests around the internal workings of its high-level operations, including the actual protocol messages those are expected to generate, and handling both successful and failure results from a simulated SP.


bnaecker:
Couple of final todos, apart from any review feedback:

- [ ] There's a big block comment in `ioloop.rs`, from the original implementation, that needs to be cut
- [ ] Remove `todo!()` in header deserialization failure path. This should take the same path that we have for similar failures elsewhere: if there's no outstanding request, drop it; if there is, bump the "faulty" message count and possibly fail the message.
- [ ] Decide on the inner / outer version number. The only practical ways to _use_ both is to try to deserialize the inner message version _only_ when deserializing the full message fails, relying on the fact that it's the first octet. That seems...maybe fine? Otherwise, just kill the two distinct versions.

Aaron-Hartwig:
>Decide on the inner / outer version number. The only practical ways to use both is to try to deserialize the inner message version only when deserializing the full message fails, relying on the fact that it's the first octet. That seems...maybe fine? Otherwise, just kill the two distinct versions.

To capture [your comment](https://github.com/oxidecomputer/hubris/pull/1239#discussion_r1154019153) from my Hubris PR with respect to inner/outer version sentiment so we don't have the same conversation in two places.

I'm okay with attempting to just grab the first byte of the inner message in the result of an inner message deserialization failure. But I guess to some extent then I wonder about the purpose of the inner protocol at all.

bnaecker:
The purpose is to support separate evolution of the header and the message body. To be clear, that's also possible with a single version and a message type that includes both the header and body. That's what we had before -- one would deserialize the whole thing, and if that failed, just deserialize the header (which included the single version number). This is  more explicit, but I think the same functionality can be achieved in either case.

bnaecker:
@jclulow @Aaron-Hartwig and I have now done extensive testing, and we've successfully used the series of PRs built on this to get the switch zone stood up on Rack 2 via `pilot`. We're going to start merging everything!

PR #80, opened 2023-04-23
Fix out-of-order parsing of module data

bnaecker:
- Reads for some kinds of complex data are split by ID. That issues SFF reads first, before CMIS, simply because that's the sort order of the `Identifier` enum. After reading, data was pushed onto a vector, which may not respect the actual module order. For example, a CMIS module appearing before an SFF would result in seeing the SFF data as applying to the CMIS. Fixed by using a map from module index to the read data, which is collected into a vector at the end.
- Apply a few clippy nits

### DIFF
--- a/controller/src/bin/xcvradm.rs
+++ b/controller/src/bin/xcvradm.rs
@@ -110,7 +110,7 @@ fn parse_transceivers(s: &str) -> Result<Transceivers, String> {
                 ));
             }
             ModuleId::from_indices(&indices)
-                .map(|p| Transceivers::Index(p))
+                .map(Transceivers::Index)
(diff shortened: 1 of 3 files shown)


bnaecker:
And just to confirm, this fixes the actual reason I noticed this:

```
BRM44220001 # /tmp/xcvradm -i sidecar0 -t1 power
Port  Power  Software-override
   1  Low    No
BRM44220001 # /tmp/xcvradm -i sidecar0 -t2 power
Port  Power  Software-override
   2  High   Yes
BRM44220001 # /tmp/xcvradm -i sidecar0 -t1,2 power
Port  Power  Software-override
   1  Low    No
   2  High   Yes
   ```

PR #113, opened 2023-06-26
Add support for ports disabled by Hubris

mkeeter:
This PR adds support for ports disabled by Hubris (see https://github.com/oxidecomputer/hubris/pull/1441)

- Adds a wider `StatusV2` type and message, which includes a bit for `DISABLED_BY_SP` (and 23 spare bits, just in case)
- Uses the `StatusV2` message in `xcvradm`
- Adds `HwError::DisabledBySp` and `HostRequest::ClearDisableLatch`

Before merging:
- Test on an actual Sidecar
- Bikeshedding about `StatusV2` (perhaps `ExtendedStatus`?  Or rename `Status` -> `HwStatus` and reclaim `Status`?), `HwError::DisabledBySp` and `HostRequest::ClearDisableLatch`, since I don't love any of those names
- Is this going to break anything that calls into `xcvradm`, since it changes types from `Status` -> `StatusV2` in a few places?

### DIFF
--- a/controller/src/controller.rs
+++ b/controller/src/controller.rs
@@ -53,7 +53,7 @@ use transceiver_messages::message::MessageBody;
 use transceiver_messages::message::MessageKind;
 pub use transceiver_messages::message::ProtocolError;
 use transceiver_messages::message::SpResponse;
-pub use transceiver_messages::message::Status;
+pub use transceiver_messages::message::StatusV2;
 pub use transceiver_messages::mgmt;
 use transceiver_messages::mgmt::sff8636;
 use transceiver_messages::mgmt::ManagementInterface;
@@ -360,7 +360,7 @@ impl Controller {
         let (powered_modules, powered_status) = filter_module_data(
             status_result.modules,
             status_result.status().iter(),
-            |_, st| st.contains(Status::POWER_GOOD | Status::ENABLED),
+            |_, st| st.contains(StatusV2::POWER_GOOD | StatusV2::ENABLED),
         );
         let unpowered_modules = status_result.modules.remove(&powered_modules);
 
@@ -368,7 +368,7 @@ impl Controller {
         // Off.
         let (in_reset, _) =
             filter_module_data(powered_modules, powered_status.iter(), |_, status| {
-                status.contains(Status::RESET)
+                status.contains(StatusV2::RESET)
             });
 
         // At this point, we have the set of `unpowered_modules` and `in_reset`.
@@ -425,7 +425,7 @@ impl Controller {
                             // Hardware is in charge, so report the state of the
                             // `LPMode` pin itself.
                             PowerControl::UseLpModePin => {
-                                let state = if status.contains(Status::LOW_POWER_MODE) {
+                                let state = if status.contains(StatusV2::LOW_POWER_MODE) {
                                     PowerState::Low
                                 } else {
                                     PowerState::High
@@ -556,15 +556,17 @@ impl Controller {
                 // module, only which hardware signals it responds to.
                 let (need_power_enabled, _) =
                     filter_module_data(modules, status.iter(), |_, st| {
-                        !st.contains(Status::POWER_GOOD | Status::ENABLED)
-                            || st.contains(Status::RESET)
+                        !st.contains(StatusV2::POWER_GOOD | StatusV2::ENABLED)
+                            || st.contains(StatusV2::RESET)
                     });
 
                 // Check for any modules which need power applied, but which do
                 // _not_ already have reset asserted. We're going to assert
                 // reset on them now, but emit a warning.
                 let (need_reset_deasserted, _) =
-                    filter_module_data(modules, status.iter(), |_, st| st.contains(Status::RESET));
+                    filter_module_data(modules, status.iter(), |_, st| {
+                        st.contains(StatusV2::RESET)
+                    });
                 let need_power_but_not_reset = need_power_enabled.remove(&need_reset_deasserted);
                 let modules = if need_power_but_not_reset.selected_transceiver_count() > 0 {
                     warn!(
@@ -918,7 +920,7 @@ impl Controller {
 
     /// Report the status of a set of transceiver modules.
     pub async fn status(&self, modules: ModuleId) -> Result<StatusResult, Error> {
-        let message = Message::from(HostRequest::Status(modules));
+        let message = Message::from(HostRequest::StatusV2(modules));
         let request = HostRpcRequest {
             header: self.next_header(MessageKind::HostRequest),
             message,
@@ -927,7 +929,7 @@ impl Controller {
         let response = self.rpc(request).await?;
         match response.message.body {
             MessageBody::SpResponse(
-                st @ SpResponse::Status {
+                st @ SpResponse::StatusV2 {
                     modules,
                     failed_modules,
                 },
@@ -935,9 +937,8 @@ impl Controller {
                 let data = response.data.expect("Existence of data checked earlier");
                 let (data, error_data) = data.split_at(st.expected_data_len().unwrap());
                 let status = data
-                    .iter()
-                    .copied()
-                    .map(|x| Status::from_bits(x).unwrap())
+                    .chunks_exact(4)
+                    .map(|x| hubpack::deserialize(x).unwrap().0)
                     .collect();
                 let failures = Self::deserialize_hw_errors(failed_modules, error_data)?;
                 Ok(StatusResult {
--- a/controller/src/controller.rs
+++ b/controller/src/controller.rs
@@ -863,6 +863,14 @@ impl Controller {
             .await
     }
 
+    /// Clears the `disabled` flag for a set of transceiver modules
+    ///
+    /// This allows them to be powered on again
+    pub async fn clear_disable_latch(&self, modules: ModuleId) -> Result<AckResult, Error> {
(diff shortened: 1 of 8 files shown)


controller/src/controller.rs
```diff
@@ -918,7 +920,7 @@ impl Controller {
 
     /// Report the status of a set of transceiver modules.
     pub async fn status(&self, modules: ModuleId) -> Result<StatusResult, Error> {
```
bnaecker, line 922: Thinking out loud, it might help the transition for this to attempt the `StatusV2` message, falling back to `Status` if that fails in a particular way. I think we can tell if a message is unsupported, though we've had few chances to test that.

messages/src/message.rs
```diff
@@ -477,6 +491,23 @@ pub enum SpResponse {
         /// The modules on which fetching the LED state failed.
         failed_modules: ModuleId,
     },
+
+    /// The extended status of a set of transceiver modules.
```
bnaecker, line 495: On the naming question, I think I prefer in order:

- Renaming `Status` to `HardwareStatus` or similar and making this `Status`.
- Keeping `Status` as is, and making this `ExtendedStatus`
- Anything else :)

messages/src/message.rs
```diff
@@ -494,6 +525,12 @@ impl SpResponse {
                 // So using `size_of` is appropriate here.
                 Some(modules.selected_transceiver_count() * core::mem::size_of::<Status>())
             }
+            SpResponse::StatusV2 { modules, .. } => {
+                // NOTE: We don't `hubpack::deserialize` the `Status` objects,
```
bnaecker, line 529: Copy-paste nit: `Status` should be `StatusV2` or whatever we name it.

messages/src/message.rs
```diff
@@ -679,6 +717,132 @@ impl SerializedSize for Status {
...
+
+        /// This module unexpectedly lost power.
+        ///
+        /// Note that this is not part of the QSFP specification, but provided
+        /// by the Sidecar board design itself.
+        const FAULT_POWER_LOST      = 0b0000_1000_0000;
+
+        /// The SP has disabled the port associated with this module
+        ///
+        /// This occurs if the module was initialized but is now NACKing I2C
```
bnaecker, line 776: This particular case is the one we've encountered so far, right? What I mean is, it seems like we should make clear this is an example, or we could make this specific case its own bit. E.g., `SP_DISABLED_DUE_TO_NACK` or something less terrible.

messages/src/message.rs
```diff
@@ -679,6 +717,132 @@ impl SerializedSize for Status {
...
+    fn expecting(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "a single u32")
+    }
+
+    fn visit_u64<E>(self, v: u64) -> Result<Self::Value, E>
+    where
+        E: SerdeError,
+    {
+        u32::try_from(v)
+            .map(StatusV2::from_bits_truncate)
```
bnaecker, line 818: Is the `from_bits_truncate` call needed here? It could just be `from_bits` since the conversion succeeded in this `map` call, right?

controller/src/controller.rs
```diff
@@ -927,17 +929,16 @@ impl Controller {
...
                 },
             ) => {
                 let data = response.data.expect("Existence of data checked earlier");
                 let (data, error_data) = data.split_at(st.expected_data_len().unwrap());
                 let status = data
-                    .iter()
-                    .copied()
-                    .map(|x| Status::from_bits(x).unwrap())
+                    .chunks_exact(4)
+                    .map(|x| hubpack::deserialize(x).unwrap().0)
```
bnaecker, line 941: Here we are using `hubpack::deserialize`, but in `impl SpResponse` we compute the size based on `size_of`. I think this is all self-consistent, because we make `SerializedSize::MAX_SIZE` equal to `size_of`, it's just confusing. I think I'd suggest we use `from_bits` here as well; or remove the comment about using `from_bits` at `message.rs:529`.

bnaecker (commented):
I think this all LGTM, thanks! I've a few suggestions about naming, but the functionality seems straightforward.

bnaecker:
> Is this going to break anything that calls into xcvradm, since it changes types from Status -> StatusV2 in a few places?

Yeah, I think if Dendrite is updated first we're gonna have a bad time. If the SP is updated first, and continues to handle `Status` and `StatusV2` messages together, then things should be OK. Dendrite uses the call to `Controller::status` all over the place, so that has to work.

The suggestion to use `StatusV2` first and fallback to `Status` would alleviate that, but it's a lot more code.

controller/src/controller.rs
mkeeter: We're using `hubpack::serialize` on the SP size, to avoid thinking about endianness, so I'll remove the comment in `message.rs`.

mkeeter: Yeah, I like the loop better; done in 514b45fbe512a91cc4f0bd68ee4287aa860256e2

messages/src/message.rs
mkeeter: Deleted this comment anyways, since we *are* using `hubpack::deserialize` here

controller/src/controller.rs
mkeeter: I decided to allow for a more graceful transition by leaving `Controller::status` as-is, and adding `Controller::extended_status` to get the wider `ExtendedStatus` message.  Dendrite can start using the latter in a later patch, instead of breaking now.

messages/src/message.rs
mkeeter: Yeah, I clarified that this is just an example (although this happens to be the only reason the SP will disable a port _right now_).

controller/src/controller.rs
bnaecker: Sounds good, the `from_bits` call ignored endianness since it's a single byte. I like it.

mkeeter:
Per chat discussion, I'm going to leave `Controller::status` returning the previous `Status` type, and add a new `Controller::extended_status` which returns `ExtendedStatus` (the 32-bit version).  This let's us switch to the new status API over time, instead of being a breaking change.

mkeeter:
Somewhat bad news: due to [a Hubris bug](https://github.com/oxidecomputer/hubris/pull/1444), the SP doesn't correctly handle messages with newer protocol versions that fail to deserialize.  Instead of replying with `MessageKind::Error`, it replies with `MessageKind::HostRequest` (copied from the incoming message), which makes `xcvradm` mad:

```
matt@niles ~ () $ ./xcvradm --log-level=trace -iaxf4 -pfe80::aa40:25ff:fe05:ff00 status
Jun 27 19:16:01.303 DEBG bound UDP socket, local_addr: Ok([::]:34099), interface: axf4
Jun 27 19:16:01.303 DEBG joined IPv6 multicast group, multicast_addr: ff02::1de:2
Jun 27 19:16:01.303 DEBG spawned IO task
Jun 27 19:16:01.303 TRCE received outgoing request, request: OutstandingHostRequest { request: HostRpcRequest { header: Header { version: 1, message_id: 0, message_kind: HostRequest }, message: Message { version: 3, body: HostRequest(ExtendedStatus(ModuleId(0xffffffff))) }, data: None }, n_retries: 0, n_error_messages: 0, response_tx: Sender { inner: Some(Inner { state: State { is_complete: false, is_closed: false, is_rx_task_set: true, is_tx_task_set: false } }) } }, task: io
Jun 27 19:16:01.303 TRCE sent outgoing request, message: Message { version: 3, body: HostRequest(ExtendedStatus(ModuleId(0xffffffff))) }, peer: [fe80::aa40:25ff:fe05:ff00%3]:11112, task: io
Jun 27 19:16:01.304 TRCE packet received, peer: [fe80::aa40:25ff:fe05:ff00%3]:11112, n_bytes: 15, task: io
Jun 27 19:16:01.304 ERRO received invalid message kind, message_kind: HostRequest, peer: [fe80::aa40:25ff:fe05:ff00%3]:11112, task: io
```

It will still work _eventually_ – after 5 retries, it falls back to calling `Controller::status`:
```
Jun 27 19:11:06.678 ERRO received too many faulty messages, limit: 5, task: io
Jun 27 19:11:06.678 WARN could not read extended status (Maximum number of faulty messages (5) reached without a response); reading status instead
```
(and this works fine)

This isn't ideal, but it's probably not worth special-casing in `xcvradm`; we should instead encourage people to update Hubris if they run into this issue.

messages/src/message.rs
mkeeter: addressed in https://github.com/oxidecomputer/transceiver-control/pull/113#issuecomment-1610017071

Aaron-Hartwig (commented):
I'll leave the final blessing to Ben, but this looks good to me.

messages/src/message.rs
```diff
@@ -679,6 +714,141 @@ impl SerializedSize for Status {
...
+
+        /// This module unexpectedly lost power.
+        ///
+        /// Note that this is not part of the QSFP specification, but provided
+        /// by the Sidecar board design itself.
+        const FAULT_POWER_LOST      = 0b0000_1000_0000;
+
+        /// The SP has disabled the port associated with this module
+        ///
+        /// For example, this could occurs if the module was initialized but is
```
bnaecker, line 776: Spelling nit, `could occurs`.

bnaecker (approved):
Looks good, thanks for the quick work!

PR #140, opened 2023-08-17
Improve reading text from stdin and help text

bnaecker:


bnaecker:
Did a bit more work here. We can now auto-detect the input format for _text_ from stdin. Strings starting with `0x` are read as hex, `0b` as binary, and anything else as decimal. There is **no default** for the input kind anymore. If it is omitted, then the text input is auto-detected. If it's provided, the input either from stdin or the file is read in that format. If the input file is specified, then the `input-kind` must also be specified. (Clap verifies that for us.)

bnaecker:
@arjenroodselaar Let me know if you have comments, or I'll just merge away.

PR #200, opened 2024-01-12
Add more `HwError` variants

Aaron-Hartwig:
Additions in support of https://github.com/oxidecomputer/hubris/issues/1562

### DIFF
--- a/messages/src/message.rs
+++ b/messages/src/message.rs
@@ -56,7 +56,12 @@ pub mod version {
 
         /// Includes `ExtendedStatus` and `clear_disabled_latch`
         pub const V3: u8 = 3;
-        pub const CURRENT: u8 = V3;
+
+        /// Additions to HwError: NoFrontIo, FrontIoNotReady, NotPresent,
+        /// NotPowered, PowerFault, NotInitialized, I2cAddressNack, I2cByteNack
+        pub const V4: u8 = 4;
+
+        pub const CURRENT: u8 = V4;
         pub const MIN: u8 = V1;
     }
     pub mod outer {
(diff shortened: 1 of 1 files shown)


Aaron-Hartwig:
I haven't been able to exhaustively test this yet as I need some remote hands for that, but here's an example of where things are headed:
```
aaron@niles ~ $ ./xcvradm -i axf6 -t 4,5,6 status
Port Status
   4 PRESENT | RESET | INTERRUPT
   5 PRESENT | ENABLED | RESET | POWER_GOOD
   6 ENABLED | RESET | INTERRUPT
aaron@niles ~ $ ./xcvradm -i axf6 -t 4,5,6 vendor-info
Port Identifier               Vendor           Part             Rev  Serial           Mfg date
Some operations failed, errors below
Port Error
   4 Hardware error accessing module 4: The module is not powered
   5 Hardware error accessing module 5: The module is not initialized
   6 Hardware error accessing module 6: The module is not present
```

messages/src/message.rs
```diff
@@ -114,7 +119,12 @@ pub enum ProtocolError {
...
 #[cfg_attr(any(test, feature = "std"), derive(thiserror::Error))]
 pub enum HwError {
-    /// The FPGA reported an I2C error
+    /// The FPGA reported an I2C error (deprecated)
+    ///
+    /// This error has been deprecated in favor of the more explicit and
+    /// informative errors below (NotPresent, NotPowered, PowerFault,
+    /// NotInitialized, I2cAddressNack, I2cByteNack). It is being maintained
+    /// for backwards compatability.
     #[cfg_attr(any(test, feature = "std"), error("FPGA reported an I2C error"))]
```
bnaecker, line 128: I think we might want to [decorate this](https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-deprecated-attribute) with `#[deprecated = "Use the more specific error variants instead"]` or similar.

bnaecker (approved):
Looks good, I think these will be very helpful up-stack! What is the plan for releasing it? Just hoping to avoid too many version-mismatch errors!

messages/src/message.rs
Aaron-Hartwig: Good call!

Aaron-Hartwig:
>What is the plan for releasing it? Just hoping to avoid too many version-mismatch errors!

Well, I think we'd like to get this into dendrite/hubris ASAP since this gives us a much richer view of what is actually happening at the transceiver module. Release 7 is shipping right now, so presumably this will get rolled into Release 8. We should maybe chat with the update folks about how to best go about this.

PR #278, opened 2024-09-18
Adds `HwError::I2cSclStretchTimeout`

Aaron-Hartwig:
We now support clock stretching when communicating with transceiver modules and that has lead to another potential error to be reported. See https://github.com/oxidecomputer/quartz/pull/210 for details.


bnaecker (approved):
🚢

PR #305, opened 2024-12-13
basic osfp support, combine the powers of reads/parse

rcgoodfellow:


decode/src/lib.rs
```diff
@@ -57,4 +57,14 @@ pub trait ParseFromModule: Sized {
 
     /// Parse the result of the above reads into `Self`.
     fn parse<'a>(id: Identifier, reads: impl Iterator<Item = &'a [u8]>) -> Result<Self, Error>;
+
+    /// Use `reads` and `parse` together to collect data from a buffer.
+    fn from_buf(id: Identifier, data: &[u8]) -> Result<Self, Error> {
```
bnaecker, line 62: How is this being or going to be used? It looks like `data` needs to already have the entire memory map of the module in order for the offsets in the reads to correctly index it. Also, this doesn't take into account the page information in the reads, so I'm not sure how this would be used to parse that spans more than one page.

bnaecker (commented):
Thanks for putting this up! I've got a few nits and one bigger question, let me know if something isn't clear.

decode/src/datapath.rs
```diff
@@ -183,7 +186,10 @@ impl ParseFromModule for Datapath {
                     MemoryRead::new(sff8636::Page::Upper(sff8636::UpperPage::new(0)?), 192, 1)?;
                 Ok(vec![tx_enable, los, cdr, compliance, extended_compliance])
             }
-            Identifier::QsfpPlusCmis | Identifier::QsfpDD => {
+            Identifier::QsfpPlusCmis
+            | Identifier::QsfpDD
+            | Identifier::Osfp8
+            | Identifier::OsfpXd => {
```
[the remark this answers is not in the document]
rcgoodfellow: The OSFP MSA states the following in the opening paragraph.

> This specification defines the electrical connectors, electrical signals and power supplies,
and mechanical and thermal requirements of the OSFP Module, connector, and cage
systems. The OSFP Management interface is described in a separate document: “Common
Management Interface Specification (CMIS)”

https://osfpmsa.org/assets/pdf/OSFP_Module_Specification_Rev5_1.pdf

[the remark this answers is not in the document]
rcgoodfellow: SFF 8024 also states in Table 3-1 that OSFP and OSFP-XD are CMIS

decode/src/lib.rs
rcgoodfellow: I'm currently using this to parse EEPROM data made available by a Linux kernel driver as a file in sysfs.

https://github.com/oxidecomputer/x2/blob/860a02811c57fcff8fa5e281c8baf788ae944838/x2-evb/src/lib.rs#L43-L53

bnaecker: I don't know the details of the Linux implementation there, but I don't understand how that can work. The datapath information is spread across many different pages, for example. How does the driver expose the idea of a page?

PR #328, opened 2025-02-08
Use a string schema for large enum types

### LINKED ISSUE #327: Enormous variants should just be schematized as strings
We have some truly huge variants in the `transceiver-decode` crate for representing things like the extended specification compliance code. These have on the order of 100 variants, but by far the most important thing about them is their display representation. I added serde attributes to serialize them all as strings, but we kept the fact that they're enums in their JSONSchema. There's no real reason for that. Anyone who reads these values through a server's API (like Dendrite's) will just treat them as read-only strings, rather than trying to act on the variant programmatically. We should just make the JSONSchema for these a bare "string", to ease downstream use of them.

bnaecker:
- Closes #327

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -191,26 +191,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "acbf1af155f9b9ef647e42cdc158db4b64a1b61f743629225fde6f3e0be2a7c7"
 
 [[package]]
-name = "const_format"
-version = "0.2.34"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


bnaecker:
I'm missing at least a few enums here: 

- `ConnectorType`
- `SffComplianceCode`
- `MediaInterfaceId`. This one should probably just be a "normal" externally-tagged enum with the stringified internal variant

ahl (approved):
Let's get that money

ahl (commented):
as discussed; I think we need to make sure the serialize/deserialize representations align with the json schema

ahl (commented):
looks good

Issue #330, opened 2025-02-08
Should implement or derive `JsonSchema` for `Status` and `ExtendedStatus`


bnaecker:
We should implement or derive JSONSchema stuff for the status bitflags. These will probably we integers, which is good enough, though that might mean downstream crates want to replace the Progenitor-implemented type with the real one. We already do implement serialization / deserialization through an integer, so that should "just work".

bnaecker:
Closed by #334

PR #334, opened 2025-02-10
Implement JSONSchema for status bit types

bnaecker:


bnaecker:
This closes #330, by delegating to the schema for the underlying integer types `u8` and `u32`. This isn't the best for human-readability, but I think we can use the `replace` directive for these types where it's useful.

ahl (approved):
seems good? I assume new uses in consumers of this crate will need them

PR #336, opened 2025-02-11
Add descriptions to types schematized as strings

bnaecker:


decode/src/datapath.rs
```diff
@@ -194,7 +194,20 @@ impl schemars::JsonSchema for EthernetComplianceCode {
     }
 
     fn json_schema(gen: &mut schemars::gen::SchemaGenerator) -> schemars::schema::Schema {
-        String::json_schema(gen)
+        let mut s = String::json_schema(gen);
```
ahl, line 197: At this point I'd just construct it my hand rather than modifying the String schema, but ... meh. either way.

decode/src/datapath.rs
```diff
@@ -194,7 +194,20 @@ impl schemars::JsonSchema for EthernetComplianceCode {
...
+        let ::schemars::schema::Schema::Object(obj) = &mut s else {
+            unreachable!();
+        };
+        obj.metadata
```
ahl, line 201: https://docs.rs/schemars/latest/schemars/schema/struct.SchemaObject.html#method.metadata does what you want here

decode/src/datapath.rs
```diff
@@ -194,7 +194,20 @@ impl schemars::JsonSchema for EthernetComplianceCode {
...
+            .get_or_insert_with(Default::default)
+            .description = Some(String::from(
+            "The Ethernet specification implemented by a module.",
+        ));
+        s
+    }
+
+    fn is_referenceable() -> bool {
```
ahl, line 209: comment?
```suggestion
    // We want this inlined--in this case, we don't expect the type to be meaningful to consumers
    fn is_referenceable() -> bool {
```

decode/src/utils.rs
```diff
@@ -184,7 +184,17 @@ macro_rules! bitfield_enum {
                 gen: &mut ::schemars::gen::SchemaGenerator
             ) -> ::schemars::schema::Schema
             {
```
ahl, line 186: all as above

decode/src/datapath.rs
bnaecker: Nice, thanks, I'm using that in a038a0b

decode/src/datapath.rs
bnaecker: Good idea, thanks. Added in a038a0b.

ahl (approved):

PR #341, opened 2025-02-22
Always attempt to read module temperature and voltage

### LINKED ISSUE #339: Not correctly decoding temperature for pre SFF-8636 rev 2.8 modules
Modules report a bunch of diagnostic monitoring data. One important one is temperature. This is required for all separable modules (all the ones we use are), but we recently found that `xcvradm` was reporting temperature as unsupported for a supposedly compliant module:

```
root@oxz_switch0:~# xcvradm -i sidecar0 -t 18 vendor-info
Port Identifier               Vendor           Part             Rev  Serial           Mfg date
  18 Qsfp28 (0x11)            Intel Corp       SPTSBP3CLCCO     01   CRCL203602ZCS    20090500
root@oxz_switch0:~# xcvradm -i sidecar0 -t 18 monitors
Port 18
         Temperature (C): --
      Supply voltage (V): --
       Avg Rx power (mW): [1.0438,1.4170,1.1306,1.0131]
            Tx bias (mA): [47.8800,47.8800,47.8800,47.8800]
           Tx power (mW): [0.8185,0.7188,0.8847,0.9508]
                   Aux 1: --
                   Aux 2: --
                   Aux 3: --
```

The support for the temperature monitor is advertised in upper page 0, byte 220, bit 5. Bit 4 advertises voltage support. In this module, both of these appear "unsupported":

```
root@oxz_switch1:~# xcvradm -i sidecar0 -t0 read-upper --sff --page 0 220 1 --binary
Port Data
   0 [0b00001100]
```

But if you read the actual registers, they sure appear to be reporting reasonable values:

```
root@oxz_switch1:~# xcvradm -i sidecar0 -t0 read-lower --sff 22 2
Port Data
   0 [0x27,0x54]
root@oxz_switch1:~# xcvradm -i sidecar0 -t0 read-lower --sff 22 2
Port Data
   0 [0x27,0x3f]
```

That shows the LSB moving around a bit, which is consistent with small (sub 1 degree) fluctuations. So what gives?

Well, if we read the spec more carefully than I did at first, it says:

> Supply voltage monitoring implemented (0b=Not implemented or pre-Rev 2.8, 1b=Implemented)

So, either temperature isn't supported, or the module is an older rev of the SFF-8636 spec. What rev is it?

```
root@oxz_switch1:~# xcvradm -i sidecar0 -t0 read-lower --sff 1 1
Port Data
   0 [0x07]
```

Which is: `SFF-8636 Rev 2.5, 2.6 and 2.7`.

Well, that explains it. We're dealing with an older module where I'm guessing these bits were reserved, and now they mean something.

One solution here is to either read the status byte, and then use that to interpret the advertisement bits. But a simpler and more robust thing would be to just read the expected bytes for both temperature and voltage, and then report them if they are something "reasonable", where reasonable probably just means not `0x00` or `0xffff`.

bnaecker:
Fixes #339

### DIFF
--- a/decode/src/monitors.rs
+++ b/decode/src/monitors.rs
@@ -228,8 +228,34 @@ const WATT_TO_MW: f32 = 1e3;
 // Conversion from Amps to milliamps, which we report.
 const AMP_TO_MA: f32 = 1e3;
 
+fn monitor_bytes_are_absurd(bytes: &[u8; 2]) -> bool {
+    bytes == &[0, 0] || bytes == &[0xff, 0xff]
+}
+
+// Try to decode the temperature, returning `None` if it appears absurd.
+//
+// "Absurd" here means all zeros or all 0xFF.
+fn decode_temperature_if_reasonable(bytes: [u8; 2]) -> Option<f32> {
+    if monitor_bytes_are_absurd(&bytes) {
+        None
+    } else {
+        Some(decode_temperature(bytes))
+    }
+}
+
 fn decode_temperature(bytes: [u8; 2]) -> f32 {
     decode_with_scale::<i16>(bytes, TEMP_RESOLUTION)
+}
+
+// Try to decode the supply voltage, return `None` if it appears absurd.
+//
+// "Absurd" here means all zeros or all 0xFF.
+fn decode_voltage_if_reasonable(bytes: [u8; 2]) -> Option<f32> {
+    if monitor_bytes_are_absurd(&bytes) {
+        None
+    } else {
+        Some(decode_supply_voltage(bytes))
+    }
 }
 
 fn decode_supply_voltage(bytes: [u8; 2]) -> f32 {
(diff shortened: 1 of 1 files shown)


decode/src/monitors.rs
```diff
@@ -277,7 +303,13 @@ impl ParseFromModule for Monitors {
     fn reads(id: Identifier) -> Result<Vec<MemoryRead>, Error> {
         match id {
             Identifier::QsfpPlusSff8636 | Identifier::Qsfp28 => {
-                // First, we'll read support for monitoring values, and the
+                // First, read the revision compliance byte. This will tell us
+                // how to handle the fact that we might be missing advertisement
+                // bits for the temperature and voltage.
+                let page = sff8636::Page::Lower;
+                let status = MemoryRead::new(page, 1, 1).unwrap();
```
Aaron-Hartwig, line 310: small nit to include an SFF-8636 table reference since we do that for all the other reads too. Looks like Table 6-3.

Aaron-Hartwig (approved):
Nice find here!

decode/src/monitors.rs
bnaecker: Done in 8f46e98

PR #342, opened 2025-02-22
Accept hex for pages, offsets, and lengths in `xcvradm`

### LINKED ISSUE #114: `xcvradm` page arguments should accept hex
The `--page` argument to `xcvradm {read,write}-upper` currently only takes a decimal argument on the CLI. We should accept hex instead, which is in line with the specs.

bnaecker:
Fixes #114


Aaron-Hartwig (approved):
Thanks for doing this!

PR #445, opened 2026-03-06
messages: more robustly handle unknown HwError's from a peer

Aaron-Hartwig:
This can happen when the controller and peer versions get out of synch, such as during an update. Current behavior is to throw a serialization error and trash the whole message. This commit allows us to complete deserialization in the event of an unknown HwError variant and display what the underlying value was which we didn't have the variant for.

### DIFF
--- a/controller/src/controller.rs
+++ b/controller/src/controller.rs
@@ -11,6 +11,7 @@ use crate::ioloop::IoLoop;
 use crate::messages::*;
 use crate::results::*;
 use crate::Error;
+use crate::MaybeUnknown;
 use crate::PowerMode;
 use crate::PowerState;
 use crate::TransceiverError;
@@ -46,6 +47,7 @@ use transceiver_messages::merge_module_data;
 use transceiver_messages::message;
 use transceiver_messages::message::Header;
 use transceiver_messages::message::HostRequest;
+use transceiver_messages::message::HwError;
 pub use transceiver_messages::message::LedState;
 use transceiver_messages::message::MacAddrResponse;
 use transceiver_messages::message::Message;
@@ -324,12 +326,22 @@ impl Controller {
         message::deserialize_hw_errors(failed_modules, data)
             .map_err(Error::from)
             .map(|hw_errors| {
+                let mut unknowns = hw_errors.unknowns.clone();
                 let errors = hw_errors
+                    .errors
                     .into_iter()
                     .zip(failed_modules.to_indices())
-                    .map(|(source, module_index)| TransceiverError::Hardware {
-                        module_index,
-                        source,
+                    .map(|(source, module_index)| {
+                        let unknown = if source == HwError::Unknown {
+                            MaybeUnknown(unknowns.pop())
+                        } else {
+                            MaybeUnknown(None)
+                        };
+                        TransceiverError::Hardware {
+                            module_index,
+                            source,
+                            unknown,
+                        }
                     })
                     .collect();
                 FailedModules {
@@ -1393,6 +1405,7 @@ mod tests {
     use crate::test_utils;
     use crate::Controller;
     use crate::Error;
+    use crate::MaybeUnknown;
     use crate::PowerMode;
     use crate::TransceiverError;
     use crate::NUM_OUTSTANDING_REQUESTS;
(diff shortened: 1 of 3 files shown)


controller/src/controller.rs
```diff
@@ -324,12 +326,22 @@ impl Controller {
...
                 let errors = hw_errors
+                    .errors
                     .into_iter()
                     .zip(failed_modules.to_indices())
-                    .map(|(source, module_index)| TransceiverError::Hardware {
-                        module_index,
-                        source,
+                    .map(|(source, module_index)| {
+                        let unknown = if source == HwError::Unknown {
+                            MaybeUnknown(unknowns.pop())
```
bnaecker: We might want to call `.pop().unwrap()`, and then wrap that in `Some(_)` again. What I'm getting at is that if we have a `source == HwError::Unknown`, we really really want there to be an entry in `unknowns` to go with it, and it's a bug if there isn't one.

bnaecker (commented):
Nice! This is going to help a lot with changes like we've made in the last 2-3 versions, adding more hardware failure information.

controller/src/lib.rs
```diff
@@ -132,15 +134,33 @@ mod probes {
...
+/// An optional unknown hardware error code, printed only when present.
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub struct MaybeUnknown(pub Option<Unknowns>);
+
+impl fmt::Display for MaybeUnknown {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        if let Some(v) = self.0 {
+            match v {
+                Unknowns::UnknownUnknown(byte) => write!(f, " (unknown: {byte:#04x})"),
+                Unknowns::KnownUnknown => Ok(()),
```
bnaecker: One more thing -- I might actually print something special here to make sure we realize this is the "known unknown" variant. I'm not sure how to do that though?

messages/src/message.rs
```diff
@@ -213,6 +216,95 @@ pub enum HwError {
...
+            errors: Vec::with_capacity(error_count),
+        };
+        for chunk in buf.chunks_exact(ITEM_SIZE) {
+            d.push(chunk)
+        }
+        Ok(d)
+    }
+
+    fn push(&mut self, chunk: &[u8]) {
+        let err = match hubpack::deserialize(chunk).map_err(|_| ProtocolError::Serialization) {
```
bnaecker: I think we can drop this `map_err()`, since we always handle the error variant anyway and ignore the contents.

controller/src/controller.rs
```diff
@@ -327,9 +328,13 @@ impl Controller {
                 let errors = hw_errors
                     .into_iter()
                     .zip(failed_modules.to_indices())
-                    .map(|(source, module_index)| TransceiverError::Hardware {
-                        module_index,
-                        source,
+                    .map(|(error, module_index)| {
+                        let (hw_err, unknown) = error;
```
bnaecker: Tiny style nit, but you can destructure this directly in the closure arguments. Something like:


```rust
.map(|((&hw_error, &unknown), module_index)| {
    TransceiverError::Hardware {
        module_index,
        source: hw_err,
        unknown: MaybeUnknown(unknown),
    }
})
```

Now that I've written that out, I'm not certain it's actually _better_ :)

bnaecker (approved):
Looks great! Thanks for the updates, I have a few minor comments, but nothing major. Merge when ready :)

controller/src/lib.rs
Aaron-Hartwig: maybe something like ` (peer sent HwError::Unknown)`? Since that is pretty explicitly what happened.

Aaron-Hartwig: And then I'll maybe adjust the other case to `(could not deserialize {byte:#04x})` to again be explicit about what happened and how this is different from the other case.

controller/src/controller.rs
Aaron-Hartwig: It needs to be:
```rust
.map(|(&(hw_error, unknown), module_index)| {
    TransceiverError::Hardware {
        module_index,
        source: hw_err,
        unknown: MaybeUnknown(unknown),
    }
})
```
Which I think is a little cleaner, particularly unpacking the pieces without another assignment is nice.

Issue #475, opened 2026-08-28
Ethernet compliance codes can be emitted as strings that we can't parse


bnaecker:
We're using `Display` and `FromStr` to read / write the Ethernet compliance codes when they appear in an API. Here are the impls:

https://github.com/oxidecomputer/transceiver-control/blob/1f94d981731f20f553e957752c5027d15be0f49d/decode/src/datapath.rs#L151-L189

These two don't match. If we get an unknown code from the module, we write it out as `Unknown (0xAA)`. But we have a wildcard match in the `FromStr` impl that will catch that and return an error. We need to parse it the same way we emit it.

PR #476, opened 2026-08-28
Ensure we can read and write unknow compliance codes

bnaecker:


bnaecker:
Fixes #475

decode/src/datapath.rs
```diff
@@ -1139,4 +1151,12 @@ mod tests {
         let code = EthernetComplianceCode::new_unchecked(0);
         assert_eq!(code.to_string(), "Unknown (0x00)");
     }
+
+    #[test]
```
jgallagher: I was going to suggest a proptest, but since this just wraps a u8, maybe it's worth an exhaustive test instead?

```rust
for i in 0..=u8::MAX {
    let code = EthernetComplianceCode::new_unchecked(i);
    let as_str = code.to_string();
    assert_eq!(code, as_str.parse().unwrap())
}
```

or somesuch

jgallagher (approved):

decode/src/datapath.rs
bnaecker: 2f8ffe5

